PR #18, opened 2025-07-30
Add structure to code.

polina-c:
1. Improve CONTRIBUTING.md
2. Organize code

Screenshot 2025-07-30 at 1 44 05 PM


pkgs/flutter_genui/lib/src/core/conversation_widget.dart
```diff
@@ -0,0 +1,69 @@
+import 'package:flutter/material.dart';
+
+import '../model/catalog.dart';
+import '../model/dynamic_ui.dart';
+import '../model/ui_models.dart';
+import '../model/chat_message.dart';
+
+class ConversationWidget extends StatelessWidget {
```
gspencergoog, line 8: This is an example of a "Custom" widget, right?

polina-c: Not sure yet. Still catching up on the code meaning. Going top down. :)

PR #20, opened 2025-07-30
Apply `dart fix` and update USAGE.md.

polina-c:


README.md
```diff
@@ -5,6 +5,7 @@ This repository is a collection of packages and applications for the Flutter Gen
 ## TODO before publishing
 
 1. Clean up Google internal information.
+2. Make colors and sizes configurable in catalog.
```
jacobsimionato: It think it is difficult to add enough configurability.

So instead, how about we just rely on the fact that developers can just provide their own catalog widgets as necessary, and completely customize them.

pkgs/spikes/fcp_client/test/core/widget_registry_test.dart
```diff
@@ -14,14 +16,15 @@ void main() {
...
-      ) =>
-          const SizedBox();
+      ) => const SizedBox();
 
-      registry.register(CatalogItem(
-        name: 'TestWidget',
-        builder: testBuilder,
-        definition: testDefinition,
-      ));
+      registry.register(
```
jacobsimionato, line 21: Oh this is new!! Cool!

README.md
polina-c: Sure.

We at least need to organize them :)

Updated the comment.

PR #36, opened 2025-08-01
Delete all the .firebase artifacts from deployment and .gitignore them

jacobsimionato:
I believe these artifacts are generated while deploying and not actually needed to run the app.


polina-c:
Can you add steps to CONTRIBUTING.md on how to configure and run the examples, for:

1. New Google developers
2. New external developers

jacobsimionato:
> Can you add steps to CONTRIBUTING.md on how to configure and run the examples, for:
> 
> 1. New Google developers
> 2. New external developers
The files I deleted here are only generated when uploading to Firebase hosting with `firebase deploy` e.g. for https://fluttergenui.web.app/

This change is unrelated to contributing. It's only needed if we want to deploy the the app to Firebase hosting, which isn't part of the contribution flow or something developers need to do. We can just document how to deploy for ourselves in the internal readme.

polina-c:
> This change is unrelated to contributing. It's only needed if we want to deploy the the app to Firebase hosting, which isn't part of the contribution flow or something developers need to do. We can just document how to deploy for ourselves in the internal readme.

How a new team member or external developer will know how to build and run the examples locally?

Now just 'flutter run' will stop working.

jacobsimionato:
> > This change is unrelated to contributing. It's only needed if we want to deploy the the app to Firebase hosting, which isn't part of the contribution flow or something developers need to do. We can just document how to deploy for ourselves in the internal readme.
> 
> How a new team member or external developer will know how to build and run the examples locally?
> 
> Now just 'flutter run' will stop working.

No, Flutter run won't stop stop working. These files are unrelated to `flutter run`. They're only generated when doing `firebase deploy`. And deleting them from version control is fine because `firebase deploy` will actually also keep working. It's just that the files will only be cached locally per user rather than checked in to version control.

PR #37, opened 2025-08-01
Replace `pub get` with `pub upgrade` to ensure fresher versions.

polina-c:
We want fresh versions locally to be in sync with what bots are running.

See internal discussion here: https://chat.google.com/room/AAQAL-XDdiM/6uwRP-aSN7U/6uwRP-aSN7U?cls=10

### DIFF
--- a/doc/CONTRIBUTING.md
+++ b/doc/CONTRIBUTING.md
@@ -30,4 +30,4 @@ See the project's `fluttergenui` details
 
 ## Shell scripts
 
-To run a script in `tool/`, open it  in VSCode and press ⇧⌘B.
+To run a script in `tool/`, open the script in VSCode and press ⇧⌘B.


tool/refresh_packages.sh
```diff
@@ -28,11 +28,11 @@ DART_PACKAGES=(
...
+    echo "--- Refreshing packages in $pkg ---"
+    (cd "$pkg" && flutter pub upgrade)
 done
 
 for pkg in "${DART_PACKAGES[@]}"; do
-    echo "--- Running dart pub get in $pkg ---"
-    (cd "$pkg" && dart pub get)
+    echo "--- Refreshing packages in $pkg ---"
+    (cd "$pkg" && dart pub upgrade)
 done
```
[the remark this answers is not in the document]
polina-c: We do not save pubspec.lock to git and we want to upgrade packages lockally continuosly to keep the versions in sync with what bots are running.

doc/CONTRIBUTING.md
```diff
@@ -25,3 +25,7 @@ Guidances:
 
 See the project's `fluttergenui` details
 [here](https://pantheon.corp.google.com/welcome?inv=1&invt=Ab4FMw&project=fluttergenui).
+
+## Shell scripts
+
+To run a script in `tool/`, open it  in VSCode and press ⇧⌘B.
```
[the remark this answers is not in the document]
polina-c: that's wrong

tool/refresh_packages.sh
```diff
@@ -28,11 +28,11 @@ DART_PACKAGES=(
 )
 
 for pkg in "${FLUTTER_PACKAGES[@]}"; do
-    echo "--- Running flutter pub get in $pkg ---"
-    (cd "$pkg" && flutter pub get)
+    echo "--- Refreshing packages in $pkg ---"
```
yjbanov, line 31: nit: "Running flutter pub get" was clearer. "Refreshing packages" is open to interpretation.

doc/CONTRIBUTING.md
```diff
@@ -25,3 +27,7 @@ Guidances:
 
 See the project's `fluttergenui` details
 [here](https://pantheon.corp.google.com/welcome?inv=1&invt=Ab4FMw&project=fluttergenui).
+
+## Shell scripts
+
+To run a script in `tool/`, open the script in VSCode and press ⇧⌘B.
```
yjbanov, line 33: Is there a Linux and Windows equivalent for this shortcut?

tool/refresh_packages.sh
```diff
@@ -28,11 +28,11 @@ DART_PACKAGES=(
...
+    (cd "$pkg" && flutter pub upgrade)
```
yjbanov, line 32: It looks like https://github.com/flutter/genui/blob/main/.github/workflows/flutter_packages.yaml will do `pub get` anyway. Should we unify how we fetch dependencies?

tool/refresh_packages.sh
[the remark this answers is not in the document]
yjbanov: `pub upgrade` does not guarantee that locally you'll get the same versions as the bots do. This is because some packages may publish new versions between two invocations of `pub upgrade`. `pubspec.lock` does guarantee it. But yes, we'd have to start checking in `pubspec.lock` into the repo to make that work.

doc/CONTRIBUTING.md
polina-c: I do not know. We will research it when we need it. There is link to doc in the task definition.

tool/refresh_packages.sh
gspencergoog: If there is no checked-in `pubspec.lock`, it is equivalent.

polina-c: Good point. Added TODO.

tool/refresh_packages.sh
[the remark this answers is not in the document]
gspencergoog: It can be done either way. I think Yegor is correct that the build is more reproducible if we check in the `pubspec.lock` files. However, it is common practice in most packages (flutter/flutter, flutter/packages, dart-lang, etc.) to not check them in, and it does avoid a bunch of churn and merging conflicts if we don't. I think we should not check them in for now. If we have problems, we can pin to a specific version in the `pubspec.yaml` as we do in other repos, or shift to checking them in.

PR #38, opened 2025-08-01
Add unit tests for `fluter_genui` package

gspencergoog:
# Description

Adds a bunch of unit tests for `flutter_genui` so we can make changes with confidence.  It also introduces a `GenerativeModelInterface` class that abstracts the generative model somewhat, although the client is still Gemini-specific.  It's just to make writing tests easier because Firebase AI Logic (unlike google_generative_ai) has no testing interface for `GenerationConfig`, and it's a final class, so it can't be mocked.

### DIFF
--- a/pkgs/flutter_genui/lib/src/ai_client/ai_client.dart
+++ b/pkgs/flutter_genui/lib/src/ai_client/ai_client.dart
@@ -15,15 +15,15 @@ import 'tools.dart';
 
 /// Defines the severity levels for logging messages within the AI client and
 /// related components.
-typedef GenerativeModelFactory = GenerativeModelInterface Function({
-  required AiClient configuration,
-  Content? systemInstruction,
-  List<Tool>? tools,
-  ToolConfig? toolConfig,
-});
+typedef GenerativeModelFactory =
+    GenerativeModelInterface Function({
+      required AiClient configuration,
+      Content? systemInstruction,
+      List<Tool>? tools,
+      ToolConfig? toolConfig,
+    });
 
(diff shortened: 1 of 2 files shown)


pkgs/flutter_genui/test/ai_client/ai_client_test.dart
```diff
@@ -0,0 +1,343 @@
...
+      ], PromptFeedback(null, null, []));
+
+      await client.generateContent<Map<String, Object?>>(
+        [],
+        Schema.object(properties: {'key': Schema.string()}),
+      );
+
+      expect(client.inputTokenUsage, 0);
+      expect(client.outputTokenUsage, 0);
+    });
```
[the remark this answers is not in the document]
gspencergoog: This is a sort-of valid point, but the `PromptFeedback` has no effect on the token count.  What would be better would be to provide a `UsageMetadata` to the response, but `UsageMetadata` is a final class with no public constructors.

pkgs/flutter_genui/test/utils.dart
```diff
@@ -0,0 +1,56 @@
+// Copyright 2025 The Flutter Authors. All rights reserved.
```
polina-c: Can we move this file to the folder test/test_infra?
I expect we will have other test helpers, and maybe goldens,
and it is good to have non-test code separately.

test_infra is standard in some Dash packages (like DevTools).

gspencergoog: Sure, done.

PR #63, opened 2025-08-04
Switch from network images to asset images.

### LINKED ISSUE #44: Make demo show images based on assets
https://github.com/flutter/genui/pull/33 adds an Image widget which can theoretically accept a URL or an asset. I've only really tested it with URL so far though. All images in the other elements now use this Image widget mechanism rather than having their own URL/asset parameter.

https://github.com/flutter/genui/pull/32 uploads images to use as assets.

The work here is to configure the images here as assets to the app, and then update the system prompt to explain to use the assets rather than URLs. The place where we currently tell it to use URLs is here: https://github.com/flutter/genui/blob/f05f255208b49ec2cb3d753566416fb849a230e0/examples/travel_app/lib/main.dart#L50

There is a long list of URLs to use here: https://github.com/flutter/genui/blob/main/examples/travel_app/lib/src/images.dart

Ideally, we will still give the LLM access to some set of images and descriptions associated with them, to help it pick assets well based on the context. There is a JSON file checked in here which contains a mapping of description to asset image name: https://github.com/flutter/genui/blob/7a54fa4474a84d4994c1b2e1f8dc987699cdede0/examples/travel_app/assets/travel_images/images.json

We could copy that into the Dart code as a constant the same as I did with the URLs, or else load it as an asset and still template it into the system prompt.

polina-c:
This is for stability of example.
Users may choose to work with their network images.

- Fixes #44


examples/travel_app/lib/main.dart
```diff
@@ -5,9 +5,10 @@ import 'package:flutter_genui/flutter_genui.dart';
 
 import 'firebase_options.dart';
 import 'src/catalog.dart';
-import 'src/images.dart';
 
-final systemPrompt =
+late final String _imagesJson;
+
+final _systemPrompt =
```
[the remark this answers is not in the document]
polina-c: false positive, because _systemPrompt is accessed after initialization

examples/travel_app/lib/src/catalog.dart
```diff
@@ -7,6 +8,17 @@ import 'catalog/options_filter_chip.dart';
...
+const imagesCatalogPath = 'assets/travel_images';
+
+Future<String> imagesCatalogJson() async {
+  var result = await rootBundle.loadString('$imagesCatalogPath/.images.json');
+  result = result.replaceAll(
+    '"image_file_name": "',
+    '"image_file_name": "$imagesCatalogPath/',
+  );
+  return result;
+}
```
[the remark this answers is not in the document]
polina-c: we have test coverage for it

examples/travel_app/lib/src/catalog.dart
```diff
@@ -7,6 +8,18 @@ import 'catalog/options_filter_chip.dart';
...
+
+Future<String> imagesCatalogJson() async {
+  var result = await rootBundle.loadString(imagesCatalogJsonFile);
+  result = result.replaceAll(
+    '"image_file_name": "',
+    '"image_file_name": "$imagesCatalogPath/',
+  );
+  return result;
+}
+
```
jacobsimionato: I think the concept of an ImageCatalog here is kind of unrelated to the idea of a catalog of widgets. So maybe this could live in a different file, e.g. test_images.dart?

examples/travel_app/pubspec.yaml
```diff
@@ -1,4 +1,4 @@
-name: genui_client
+name: travel_app
```
jacobsimionato, line 5: Thanks for fixing this!!!

examples/travel_app/lib/src/catalog.dart
polina-c: Added '@visibleForTesting' to the path constants.

However, imagesCatalogJson() is what is executed to provide images to prompt.

I would suggest that: UI catalog = widgets + images.

polina-c:
> Hey nice work! Do the images seem to load successfully from the assets?

Yes, travel app works!

examples/travel_app/lib/src/catalog.dart
jacobsimionato: I think the use of images and widgets here is quite different, because the widget catalog is something that developers will definitely replicate in a real app. The catalog of images here would not exist in a real app - this is just a test utility to give the LLM some images to use in the absence of "real" images which would be fetched from some tool that calls a database etc. That's what my hotel_database change was about - give the LLM a tool to fetch data, which would include image references (probably URLs).

So I like the idea of separating this into catalog.dart which developers can branch and modify, and something like test_images.dart or demo_images.dart which is clearly marked as a demo-only workaround that developers will not try to replicate in a real app. We don't want developers to think that they need to declare all images that their app can use in advance, because that won't be possible for a real use case.

polina-c: I partially agree.

Yes, developers will need to provide some way to fetch images. 

First, I think this 'some way' is part of the UI catalog definition. Do you think differently?

Second, I think, this 'some way' may be combination of:
1. An MCP service (is it how we plan to shape the database?)
2. Hard coded assets or network images (like [company logo in travel app](https://github.com/flutter/genui/blob/main/pkgs/spikes/travel_app_hardcoded/assets/agent_icon.png) or small set of images of garden tools that the company sells)
3. Some other way

At the moment we provide just #2, and we have plans to implement #1. We can add a note to 'USAGE.md', explaining that the database possibility is coming soon. And, I added prefix 'asset' to specify that there can be other options to provide images. Thoughts?

jacobsimionato: Hey I expect pretty much all images will come from tool calls, potentially using MCP to expose the tools or not. I expect company logos and other assets are more likely to be baked into the CatalogItems.

But this is just a first pass on the app and I don't want to keep getting in the way, so let's stick with this.

polina-c: It is good to exchange opinions and align on mental models.
I see what you mean.  Thank you for explanations.

Moved to asset_images.dart.

PR #72, opened 2025-08-05
Refactor SDK logic and improve logging of LLM inferences

jacobsimionato:
These were some first steps I made towards redoing the event logic before I had to stop working today. They're a bit scattered!

- Rename `GenerativeModelWrapper` to `FirebaseAiGenerativeModel` seeing as this is a Firebase AI logic-specific implementation of an interface.
- Rename `DynamicUi` to `SurfaceWidget` seeing as this widget is specifically for displaying an individual surface.
- Fix a bug in `ai_client.dart` that was causing `capturedResult` to be pulled incorrectly resulting in an infinite loop of inferences.
- Refactor `GenerativeModelInterface` to remove unused parameters.

### DIFF
--- a/pkgs/flutter_genui/lib/src/ai_client/ai_client.dart
+++ b/pkgs/flutter_genui/lib/src/ai_client/ai_client.dart
@@ -449,8 +449,9 @@ $concatenatedContents
 With functions:
   '${allowedFunctionNames.join(', ')}',
   ''');
-      DateTime startTime = DateTime.now();
+      final inferenceStartTime = DateTime.now();
       final response = await model.generateContent(contents);
+      final elapsed = DateTime.now().difference(inferenceStartTime);
 
       // If the generate call succeeds, we need to reset the delay for the next
       // retry. If the generate call throws, this won't get called, and the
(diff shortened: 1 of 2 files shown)


polina-c:
It is very rare that widget names in Flutter contain the word 'Widget'. Instead they contain some word that means something visual, like 'Button' or 'View'.

How about:

* Surface
* SurfaceView
* EphemeralView
* something else?

gspencergoog:
> It is very rare that widget names in Flutter contain the word 'Widget'. Instead they contain some word that means something visual, like 'Button' or 'View'.

`DynamicSurface`?

pkgs/flutter_genui/lib/src/ai_client/ai_client.dart
```diff
@@ -275,7 +275,7 @@ class AiClient implements LlmConnection {
     List<Tool>? tools,
     ToolConfig? toolConfig,
   }) {
-    return GenerativeModelWrapper(
```
gspencergoog, line 278: I named that so that it could grow into the interface to any generative model.  I know it is still specific to Firebase AI Logic right now, but eventually it shouldn't be.

PR #78, opened 2025-08-05
Update/fix/add copyright notices

gspencergoog:
# Description

This PR updates/adds/normalizes all copyright notices on all "source code" files in the repo. No code was harmed in the making of this PR.

(I used a tool I wrote a long time ago that updates copyright notices for the Flutter repo.)


gspencergoog:
> However, I've noticed that in a few files, the new copyright header was added without removing the old one, leading to duplicate headers. I've left a comment on one of the affected files with a suggestion for a fix that should be applied to all similar cases.

This is simply wrong.  I've manually checked every one of the files.  It also never added a comment.

Issue #43, opened 2025-08-01
LLM inferences use correct triggering, context, and parallelism to mantain a conversational flow


polina-c:
# Task

This task is to tweak our use of event triggering, context, paralellism, system prompt, loading spinners etc to try to make the travel app work for typical happy path use cases and follow a conversational flow.

Video of an example journey:
https://github.com/user-attachments/assets/cecd5404-e359-4a3f-b946-014d624ebc96

Currently, when you have an example conversation with the LLM, for an example prompt like "trip to mexico", it feels kind of weird. The issues I see are:
- Inferences trigger when updating filter chips, but before clicking "find my trip". I haven't finished choosing my options yet.
- When you then update the options *after* an inference is already in progress, I believe it spins up another parallel inference with results that clobber the original inference.
  - Potentially this could be solved by making the chips enter a "loading state" immediately after pressing "find my trip", and then when the LLM response comes back, we can reinstate the chips to allow for future turns.
- When the results come back, the first ones failed for some reason. Not sure why yet!
- When future results come back, it preserves the filter chips, as requested by the system prompt. However, the state and layout is not consistent. The options I chose are lost, and it makes random changes to the configuration, e.g. using different emoji etc for the chip icons. I think this could be solved by ensuring that the current full surface contents is always included in the inference request, and there is something in the SI to suggest that the LLM should preserve parts of the UI exactly as they are when it makes sense.

Things that worked well in this video that we should ensure are consistent:
- The trip is updated in-place in the same turn when the filters were changed. I find this isn't always the case.
- When I made a followup query via trailhead about resorts in Riviera Maya, the results were shown in a new surface below. Sometimes, I find it instead replaces existing trips that I want to preserve.

PR #80, opened 2025-08-06
Supply current UI state in the prompt as context

gspencergoog:
# Description

This adds the current UI state to the prompt so that the LLM knows what the user is seeing.

# Related Issues

- Related to #43

### DIFF
--- a/pkgs/flutter_genui/lib/src/ai_client/ai_client.dart
+++ b/pkgs/flutter_genui/lib/src/ai_client/ai_client.dart
@@ -80,8 +80,8 @@ class AiClient implements LlmConnection {
     this.tools = const <AiTool>[],
     this.outputToolName = 'provideFinalOutput',
     String? systemInstruction,
-  })  : _systemInstruction = systemInstruction,
-        model = ValueNotifier(model) {
+  }) : _systemInstruction = systemInstruction,
+       model = ValueNotifier(model) {
     final duplicateToolNames = tools.map((t) => t.name).toSet();
     if (duplicateToolNames.length != tools.length) {
       final duplicateTools = tools.where((t) {
@@ -108,8 +108,8 @@ class AiClient implements LlmConnection {
     this.tools = const <AiTool>[],
     this.outputToolName = 'provideFinalOutput',
     String? systemInstruction,
-  })  : _systemInstruction = systemInstruction,
-        model = ValueNotifier(model) {
+  }) : _systemInstruction = systemInstruction,
+       model = ValueNotifier(model) {
     final duplicateToolNames = tools.map((t) => t.name).toSet();
     if (duplicateToolNames.length != tools.length) {
       final duplicateTools = tools.where((t) {
@@ -431,8 +431,9 @@ class AiClient implements LlmConnection {
 
     final model = modelCreator(
       configuration: this,
-      systemInstruction:
-          _systemInstruction == null ? null : Content.system(_systemInstruction!),
+      systemInstruction: _systemInstruction == null
+          ? null
+          : Content.system(_systemInstruction),
       tools: generativeAiTools,
       toolConfig: ToolConfig(
         functionCallingConfig: FunctionCallingConfig.any(
--- a/pkgs/spikes/usage_test/lib/main.dart
+++ b/pkgs/spikes/usage_test/lib/main.dart
@@ -35,8 +35,9 @@ class MyHomePage extends StatefulWidget {
 }
 
 class _MyHomePageState extends State<MyHomePage> {
-  final GenUiManager _genUiManager =
-      GenUiManager.conversation(llmConnection: AiClient());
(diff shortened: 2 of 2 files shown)


pkgs/flutter_genui/lib/src/core/genui_manager.dart
```diff
@@ -106,17 +108,46 @@ class GenUiManager {
...
+  }
+
   Future<void> _generateAndSendResponse({
     required List<Content> conversation,
   }) async {
     _outstandingRequests++;
     _loadingStreamController.add(true);
     try {
+      final uiState = _getCurrentUiState();
+      final tempConversation = [
```
jacobsimionato, line 126: Maybe the simpler way to model this could be to pass the entire conversation including user prompts and generated UI, exactly as the user sees it, as the first part of the conversation. And then after that, put in the events that have occured on this specific surface.

E.g. an example might be like:

User prompt: "Trip to dubai"
LLM response: Generated UI JSON
FunctionCall: Choose trip type: adventure
FunctionCall: Press submit button

I think that better represents the state of the system as the user sees it. WDYT? Hard to know if it will actually make behavior any better.

jacobsimionato: Also, side comment: I wonder how functionResponse vs just Content.text compares for representing UI events.

gspencergoog: > Also, side comment: I wonder how functionResponse vs just Content.text compares for representing UI events.

Content.text is inherently a user message, and I wanted to use the fake function responses because then the LLM doesn't look at them as instructions, but more like context. How much does it affect quality?  Who knows?

gspencergoog: > FunctionCall: Choose trip type: adventure
> FunctionCall: Press submit button

Do you mean these as literally what gets sent (as in there's something summarizing/English-izing the UI events)?  I'm assuming no, but correct me if I'm wrong.

I think that's basically what we have now (but maybe I'm wrong?): the function call responses are how we're sending the events.  What would you do differently?

pkgs/flutter_genui/lib/src/core/genui_manager.dart
```diff
@@ -106,17 +108,46 @@ class GenUiManager {
...
       final response = await llmConnection.generateContent(
-        conversation,
+        tempConversation,
         outputSchema,
         systemInstruction: Content.system(instruction),
       );
+
+      // Copy the tool-call-and-response history from the temporary
+      // conversation back to the original.
+      if (tempConversation.length > conversation.length + 1) {
```
jacobsimionato, line 141: I don't quite understand this  - aren't the tool responses already in "conversation"? And isn't tempConversation.length always equal to conversation.length + 1 ?

pkgs/flutter_genui/lib/src/core/genui_manager.dart
jacobsimionato: Sorry, my two messages created confusion. I think what you said above answers my second question.

But re the first question:

No, sorry I didn't mean literally sending the events as english text, I was just trying to suggest ordering all the information in the list of Content so that it more closely matches what the user is looking at in the UI.

So imagine a conversation like:
1. User says "trip to london"
2. Llm show filter chips for London
3. User picks filters
4. LLM shows itinerary for london, with filter chips above it
5. User says "Now help me plan a stopover in Dubai"
6. LLM shows new filter chips in a new turn below
7. User picks filters for budget: luxury

If I'm reading the code, right, I think what we're sending as context to the LLM now is something like:

- User prompt: "trip to london"
- User prompt: "Now help me plan a stopover in Dubai"
- FunctionCall: Trip type: luxury
- FunctionCall: Press submit button
- Text summary of the current UI state for all gen UIs, which includes:
  - Filter chips and itinerary for london
  - Filter chips for Dubai
 
 I think that could be confusing, because it's grouping things by content type, rather than how the user sees it, and what happened chronologically. I think it should be:
 
- User prompt: "trip to london"
- UI description: Filter chips and itinerary for london - maybe this should be `Content.model([TextPart(genUiJson)])` so the model can tell that it generated this?
- User prompt: "Now help me plan a stopover in Dubai"
- UI description: Filter chips for dubai
- FunctionCall: Trip type: luxury
- FunctionCall: Press submit button

I've had some experiences where when you continue the conversation, the LLM starts being confused about which question it is trying to answer, because it can't see the existing back and forth between the model and user, so I'm trying to prevent this.

gspencergoog: Oh, yes.  That makes sense now.  Sure, let's try and reorder things so that's how it works.  I'm done for today, but maybe you can take this PR and start from there? If not, I can take a look at that tomorrow.

gspencergoog:
Closing in favor of #87

PR #81, opened 2025-08-06
fix some analyzer infos

gspencergoog:
# Description

Just a drive by fixing some analyzer infos.  I'm not sure why the CI didn't pick this up... have to look into that.


gspencergoog:
This doesn't yet fix the CI, but it does fix a typo there that I made a few PRs back.

Issue #46, opened 2025-08-02
Make sure progress indicator shows up in right time and place


polina-c:
The progress indicator should show up where the mocks specify, and whenever an inference is in progress.

PR #93, opened 2025-08-06
Reduce boilerplate by distinguishing two types of GenUI: chat and flexible.

polina-c:
Fixes https://github.com/flutter/genui/issues/46
Fixes https://github.com/flutter/genui/issues/101

### DIFF
--- a/.gemini/GEMINI.md
+++ b/.gemini/GEMINI.md
@@ -1,4 +1,6 @@
-# CatalogItem
+# Instructions for code review
+
+## CatalogItem
 
 A CatalogItem is an object which represents a widget that can be instantiated by an LLM. It centralizes the widgetBuilder function for the widget along with the data schema and name.
 
@@ -47,3 +49,8 @@ The folder `spikes` contains experiments and proof of concepts,
 that does not have to be of good quality.
 
 Skip this folder when reviewing code.
+
+## Draft pull requests
+
+Do not review pull requests, when they are in draft state.
+Wait them to be ready for review.


examples/generic_chat/lib/main.dart
```diff
@@ -10,13 +10,20 @@ import 'package:flutter_genui/flutter_genui.dart';
 import 'firebase_options.dart';
 
 final systemPrompt =
-    '''You are a helpful assistant who figures out what the user wants to do and then helps suggest options so they can develop a plan and find relevant information.
+    '''You are a helpful assistant who figures out what the user wants to do
```
gspencergoog: Wrapping these makes them not flow properly for me: you have a different line wrap algorithm than I do.  Not wrapping them, and having local wrap on, means that they wrap correctly for both of us, regardless of our wrap algorithm differences.

For example, without wrapping:
image

With your wrapping:
image

Not wrapping also saves a few tokens.

polina-c: Oh, I did not know. Thanks.

Fixed.

pkgs/flutter_genui/lib/src/ai_client/gemini_ai_client.dart
```diff
@@ -15,6 +15,19 @@ import 'gemini_schema_adapter.dart';
 import 'generative_model_interface.dart';
 import 'tools.dart';
 
+const _chatPromptPrefix =
+    '''You are a helpful assistant who figures out what the user wants to do
```
gspencergoog: Can you unwrap this one too?

PR #100, opened 2025-08-07
Fix catalog to not default to empty, because apps fail with empty catalog.

polina-c:


gspencergoog:
Merged so I can handle conflicts in my other PRs.

PR #121, opened 2025-08-13
Fix schema adapter to propagate "description" field and add a test

jacobsimionato:
The behavior of the travel app was starting to feel a but funky recently. This was because the descriptions in the schema which have hints like "always put filter chips in a filter chip group" were being dropped with the migration to the new Schema library. This reinstates them! I believe this was also the reason the images stopped working suddenly. We had descriptions which were explaining to use an image widget.


pkgs/flutter_genui/test/ai_client/gemini_schema_adapter_test.dart
gspencergoog, line 1: I think there's already a `test/ai_client/schema_adapter_test.dart` We should combine them (and probably keep your name, since it's more consistent).

gspencergoog:
Thanks!  I noticed this today too, but didn't get around to fixing it yet.

pkgs/flutter_genui/test/ai_client/gemini_schema_adapter_test.dart
jacobsimionato: Ahhh sorry I missed this! I (well, Gemini) merged them now.

PR #122, opened 2025-08-13
Add facade.

polina-c:


pkgs/flutter_genui/lib/src/facade/firebase_ai.dart
```diff
@@ -0,0 +1,42 @@
+// Copyright 2025 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:firebase_ai/firebase_ai.dart';
```
yjbanov: We can't have a direct dependency on `package:firebase_ai`, because the genui package is meant to be agnostic of the AI backend, right? Can we instead have the core package provide some kind of abstraction that a separate package would implement using `package:firebase_ai`? Then some other package could implement it using ADK, or LangChain, or something else. All without the core genui package depending on all of the AI providers.

polina-c: Yes, we can provide abstraction.

However, I want to challenge this "genui package is meant to be agnostic of the AI backend".

We observe big difference in ai model's APIs. And, their input and output became not just string, but something more structured.

They all are very different and they will continue popping out and enhancing, and we want to take advantage of all their features without being limited by any cross-model consistency obligations.

So, I suggest having a separate facade for each type of ai.

And it may happen, that they will be different by just constructor, so it will be easy for users to switch between them.

In our core library we maybe will find it convenient to have some agnostic parts. But we want to be free to modify it without breaking API.

How does it sound?

@gspencergoog

polina-c: We can have generic facade as well, but it will be less functional and less reliable.

polina-c: Also, when we provide API, that is tightly coupled with a technology, it is easier for users to integrate with it.

gspencergoog: > Can we instead have the core package provide some kind of abstraction that a separate package would implement using package:firebase_ai? 

We actually already have that abstraction, I spent part of last week adding it, but we haven't yet factored it out into a separate package (or packages). To truly be agnostic, it probably needs to be some kind of federated package so you can depend on the ones you want to use and don't have to implement them yourself.

I factored the abstraction in such a way that the `AiClient` API is agnostic of the underlying API, and the individual implementations adapt the input to their API.

See https://github.com/flutter/genui/tree/main/pkgs/flutter_genui/lib/src/ai_client 

There's currently only a Gemini implementation, and it adapts the generic JSON schema and `ChatMessage` abstractions to its own (`firebase_ai`) implementations, but the `ChatMessage`, `AiTool` and `Schema` classes are independent of which LLM is used.

polina-c: The question is: do we want to have facade per model or we want user to do extra effort to adapt the model they use to genui?

Issue #114, opened 2025-08-12
Implement facade for simple chat.


sethladd:
TLDR: Wish for an easy way to add GenUI to my existing Chat App (Firebase AI Logic first)
TLDR: Non-invasive way to integrate the Gen UI surface widgets and state with an existing chat UI implementation

I have an existing app which already has a chat experience. For example, https://github.com/sethladd/throwaway-genui-test

I want to add GenUI to my chat.

I think we need additional APIs to make it more straight forward, assuming the user is already using Firebase AI Logic and already has a chat feature.

PR #127, opened 2025-08-13
Restructure Gen UI manager interactions with widgets to allow for custom conversation widget implementations

jacobsimionato:
This change makes it so that GenUiManager exposes all the data needed to render a conversation, and the chat widget simply accepts it in the constructor. That way, apps with existing chat UIs can act as the conversation widget implementation and integrate directly with GenUiManager.

This change also removes some of the complexity from GenUiManager around the support of multiple conversation UI implementations.

This is intended to help https://github.com/flutter/genui/issues/114


polina-c:
Let's discuss on standup. It may be there is simpler API.

Issue #66, opened 2025-08-04
Remove firebase config from codebase and update readme on how to create it


jacobsimionato:
We should remove GoogleServices.json and firebase_options.dart and .firebaserc from version control.

PR #137, opened 2025-08-14
Clean flutter and firebase configuration.

polina-c:
Used this as starting point: https://github.com/flutter/genui/compare/main...remove_firebase_config_2

Fixes https://github.com/flutter/genui/issues/66


polina-c:
> It would be nice if we had a script that would configure firebase for local running so it would be easy to set up, but we can do that in another PR.

Yes, I plan to do it in next PR.

PR #138, opened 2025-08-14
Rename .images.json to images.json to fix deployed web app

jacobsimionato:
Using the name with '.' in it seems to break the firebase deployment e.g. https://fluttergenui.web.app/.

When I was loading the URL above, I would get:
```
main.dart.js:30598 Error while trying to load an asset: Flutter Web engine failed to fetch "assets/assets/travel_images/.images.json". HTTP request succeeded, but the server responded with HTTP status 404.
```

It might be that the deployment script ignores these files or the web server does so. But either way, it seems simplest to just use a filename without the dot in it.


jacobsimionato:
I've already redeployed the app with this PR added, so you won't see the error any more.

examples/travel_app/lib/src/asset_images.dart
```diff
@@ -4,7 +4,7 @@ import 'package:flutter/widgets.dart';
 @visibleForTesting
 const assetImageCatalogPath = 'assets/travel_images';
 @visibleForTesting
-const assetImageCatalogJsonFile = '$assetImageCatalogPath/.images.json';
+const assetImageCatalogJsonFile = '$assetImageCatalogPath/images.json';
```
polina-c: Can we rename it to _images.json, to keep on top of the list?

jacobsimionato: Good idea! Done.

PR #145, opened 2025-08-15
Small updates to Dart Schema Builder

gspencergoog:
# Description

When I was playing with the FCP prototype, I found some small issues in the schema builder that I corrected.


gspencergoog:
> My main concern is the removal of validation logic for string properties like minLength, maxLength, and pattern, which seems to be a critical regression. I've also provided a suggestion to further improve the email validation regex.

The validation responsibility was moved from the StringSchema class into lib/src/schema_validation.dart, inside the validateTypeSpecificKeywords method. This method is called as part of the overall schema validation process. So, even though the logic is no longer in  string_schema.dart, it's still being applied correctly.

PR #153, opened 2025-08-19
Remove copyright header from generated files

jacobsimionato:
Without this change, I see the following in my git repo, because when Flutter regenerates these files from time to time, the copyright header gets removed:

```
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   examples/travel_app/macos/Flutter/GeneratedPluginRegistrant.swift
	modified:   examples/travel_app/windows/flutter/generated_plugin_registrant.cc
	modified:   examples/travel_app/windows/flutter/generated_plugin_registrant.h
```

I think we can remove the header from the generated file for practical purposes. We already don't have a copyright header on generated files such as pubspec.lock.

### DIFF
--- a/examples/travel_app/macos/Flutter/GeneratedPluginRegistrant.swift
+++ b/examples/travel_app/macos/Flutter/GeneratedPluginRegistrant.swift
@@ -1,7 +1,3 @@
-// Copyright 2025 The Flutter Authors. All rights reserved.
-// Use of this source code is governed by a BSD-style license that can be
-// found in the LICENSE file.
-
 //
 //  Generated file. Do not edit.
 //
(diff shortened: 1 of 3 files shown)


gspencergoog:
But the reality is that we need to have Copyright headers on all files with source in them.  For instance, we have them on generated files in the Flutter framework repo.

It's trivial to run `dart run tool/fix_copyright/bin/fix_copyright.dart --force` to fix them all. (and we could make it into a simple bash script that runs that to make it easier to type). We also don't regenerate these all that often.

jacobsimionato:
Understood! Apologies for the noise. I'll drop this.

PR #156, opened 2025-08-19
Define new API for GenUiClient.

polina-c:
Next step: define the facade SimpleChatGenUi that will use this API.

### DIFF
--- a/pkgs/flutter_genui/lib/flutter_genui.dart
+++ b/pkgs/flutter_genui/lib/flutter_genui.dart
@@ -13,6 +13,7 @@ export 'src/catalog/core_widgets/text.dart';
 export 'src/catalog/core_widgets/text_field.dart';
 export 'src/core/core_catalog.dart';
 export 'src/core/genui_manager.dart';
+export 'src/core/new_genui_manager.dart';
 export 'src/core/ui_event_manager.dart';
 export 'src/model/catalog.dart';
 export 'src/model/catalog_item.dart';


pkgs/flutter_genui/lib/src/core/genui_manager.dart
```diff
@@ -1,287 +1,48 @@
...
-    _init(catalog);
-    _chatController = GenUiChatController();
-    loadingStream.listen((bool data) {
-      if (data) {
-        _chatController?.setAiRequestSent();
-      }
-    });
+  }) : _surfaces = surfaces {
+    this.catalog = catalog ?? coreCatalog;
   }
```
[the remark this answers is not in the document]
polina-c: false positive - catalog has to be const for this

examples/minimal_genui/lib/main.dart
```diff
@@ -47,7 +47,7 @@ class MyHomePage extends StatefulWidget {
 }
 
 class _MyHomePageState extends State<MyHomePage> {
-  final GenUiManager _genUiManager = GenUiManager.chat(
+  final OldGenUiManager _genUiManager = OldGenUiManager.chat(
```
gspencergoog: Instead of renaming this now, why not name the new one `NewGenUIManager` for now and rename it `GenUIManager` when you remove the old one?

polina-c: done!

pkgs/flutter_genui/lib/src/core/new_genui_manager.dart
```diff
@@ -0,0 +1,91 @@
...
+  /// If the surface is not defined by AI yet, will use default builder.
+  ///
+  /// If [defaultBuilder] is not provided, `SizedBox.shrink()` will be rendered.
+  ///
+  /// If the surface with [surfaceId] does not exist in [surfaces],
+  /// will throw an error.
+  Widget build({
+    required BuildContext context,
+    required String surfaceId,
+    WidgetBuilder? defaultBuilder,
```
jacobsimionato, line 70: It seem as if this is an alternative type of API to the surfaceUpdates below.

As we discussed in the meeting yesterday, rather than having all the widget rendering logic centralized inside Gen UI Manager, how about we make it more modular and familiar by instead having developers directly create our existing SurfaceWidget ( https://github.com/flutter/genui/blob/65ea5a1e96bbc017b6f2e3bb20080c24ef0ce1ae/pkgs/flutter_genui/lib/src/model/surface_widget.dart#L4) to do the rendering for a surface, and let it subscribe to data from the manager.

E.g. we update SurfaceWidget so that it subscribes to a stream of UiDefinitions etc so it can update automatically.

The developer calls something like `SurfaceWidget(surfaceId: "some id", host: _genUiManager)`.

I proposed something a bit like that in my PR here:
https://github.com/flutter/genui/blob/4354084580e4bb574b44d9a675f133d6bd78a865/pkgs/flutter_genui/lib/src/model/surface_widget.dart

pkgs/flutter_genui/lib/src/core/new_genui_manager.dart
```diff
@@ -0,0 +1,91 @@
...
+  /// The warning message.
+  String get message;
+}
+
+// TODO: rename to GenUiManager after implementing.
+class NewGenUiManager {
+  NewGenUiManager({
+    required this.aiClient,
+    required GenUiSurfaces surfaces,
+    required this.generalPrompt,
```
jacobsimionato, line 33: Let's remove this from the API - developers should set this directly on their AI integration

pkgs/flutter_genui/lib/src/core/new_genui_manager.dart
```diff
@@ -0,0 +1,91 @@
...
+abstract class GenUiWarning {
+  /// The warning message.
+  String get message;
+}
+
+// TODO: rename to GenUiManager after implementing.
+class NewGenUiManager {
+  NewGenUiManager({
+    required this.aiClient,
+    required GenUiSurfaces surfaces,
```
jacobsimionato, line 32: If this is set initially, I'm wondering how Gen UI will create new surfaces in a conversational style. In a conversation, we currently model it as one surface per turn. I think developers will need a callback for when surfaces are created (or later deleted) to enable this.

Also, if we are supporting the scenario where there is a linear conversation of surfaces, but also additional pinned surfaces, who decides how that is represented in the conversation context? We currently maintain this in _chatHistory in Gen UI manager.

I think that with this new setup, the Gen UI manager should be super flexible and perhaps just ask the layer above to represent the current state as a conversation data structure and give it to us

pkgs/flutter_genui/lib/src/core/new_genui_manager.dart
polina-c: There is no big difference between these two calls:

- genui.build(surfaceId)
- SurfaceWidget(genui, surfaceId)

But, yes, SurfaceWidget seem to be more familiar.

@gspencergoog , what do you think?

pkgs/flutter_genui/lib/src/core/new_genui_manager.dart
polina-c: The suggestion is that surfaces can be updated any moment and influence any further request. I do not see why developer would need callback. Just setter seem to be enough.

In API there is no differentiation if a surface is pinned or not pinned. There is just current snapshot of surfaces and explanation how they work.

PR #161, opened 2025-08-20
Create fix_copyright.sh

polina-c:


tool/fix_copyright.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# Fast fail the script on failures.
+set -ex
+
+# The directory that this script is located in.
+TOOL_DIR=$(dirname "$0")
+
+(
+  cd "$TOOL_DIR/.."
+  dart tool/fix_copyright/bin/fix_copyright.dart --year "$(date +%Y)" --force
```
gspencergoog: The year should be fixed at 2025, since that is when the project started.  For example, the Flutter framework files all have 2014, even the new ones.

tool/fix_copyright.sh
```diff
@@ -0,0 +1,17 @@
...
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# Fixes copyright headers to make bots happy.
+
+# Fast fail the script on failures.
+set -ex
+
+# The directory that this script is located in.
+TOOL_DIR=$(dirname "$0")
```
gspencergoog: Use `${BASH_SOURCE[0]}` instead of `$0`.  You might also use realpath to expand links.

```suggestion
TOOL_DIR=$(dirname $(realpath "${BASH_SOURCE[0]}"))
```

tool/fix_copyright.sh
gspencergoog, line 1: Make sure this has had `chmod +x` run on it before adding to git so it keeps the mode.

tool/fix_copyright.sh
```diff
@@ -0,0 +1,17 @@
...
+
+(
```
gspencergoog: Since this is inside of a script, the parens aren't necessary (and just make it ever so slightly slower, since it has to fork a new shell).

tool/fix_copyright.sh
```diff
@@ -0,0 +1,17 @@
...
+  cd "$TOOL_DIR/.."
```
gspencergoog: I actually don't think you want to do this.  Instead, I'd just pass the full path to the Dart script to dart and give it all the command line arguments that were passed in, using `"$@"` so that people can run the script over just part of the tree. It will use the cwd for a starting point then.

tool/fix_copyright.sh
polina-c: This is not always true: devtools puts year of file creation.
But I like when all files have the same year. Fixed.

tool/fix_copyright.sh
```diff
@@ -0,0 +1,15 @@
...
+# Fixes copyright headers to make bots happy.
+
+# Fast fail the script on failures.
+set -ex
+
+# The directory that this script is located in.
+TOOL_DIR=$(dirname $(realpath "${BASH_SOURCE[0]}"))
+
+# The year is hardcoded to 2025, year of project creation.
+dart "$TOOL_DIR/fix_copyright/bin/fix_copyright.dart" --year 2025 --force
```
gspencergoog: I'd add `$@` at the end, just in case people want to add more args on the fly (like a particular file or dir)

```suggestion
dart "$TOOL_DIR/fix_copyright/bin/fix_copyright.dart" --year 2025 --force "$@"
```

PR #186, opened 2025-08-25
Alias Json type, without exposing in API.

polina-c:


pkgs/flutter_genui/lib/src/primitives/simple_items.dart
```diff
@@ -0,0 +1,5 @@
+// Copyright 2025 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+typedef Json = Map<String, Object?>;
```
gspencergoog: I don't think this should be called just "Json", since it's not a JSON string or native JSON object. It's a map that can be used for JSON, so maybe `JsonMap`?

Issue #131, opened 2025-08-14
Redesign the core catalog to provide a usable out-of-the-box experience


jacobsimionato:
The core catalog currently is:

```dart
final coreCatalog = Catalog([
  elevatedButtonCatalogItem,
  columnCatalogItem,
  text,
  checkboxGroup,
  radioGroup,
  textField,
  image,
]);
```

If you try to use them in a basic app, it doesn't provide an inspiring experience.

Example:
<img width="1206" height="2622" alt="Image" src="https://github.com/user-attachments/assets/55ece2b2-5ecf-4f77-888b-2e29bcf94947" />

Note the text box is weird without being in a card etc, and there is no submit button.

Some issues are:
- Most of the input elements use UiChangeEvent which doesn't trigger an inference, so they need to be paired with a "Submit button". E.g. when I get a UI with a text box, I have no way to submit my input. We should add descriptions to the schemas of these elements, e.g. CheckboxGroup, TextField to explain that they should only be used with a button to submit.
- Lack of output elements beyond text e.g. Card, List etc. Without these, it's hard for the UI to make sense. We should have guidance e.g. to bundle related sets of input elements inside a Card with a submit button.

The goal here is that someone can do a minimal integration and see some basic UI appear that at least makes sense, even if it isn't visually polished.

PR #187, opened 2025-08-25
Add field "exampleData" to catalog item and create gallery to test the catlogs.

polina-c:
Contributes to https://github.com/flutter/genui/issues/131

Also:

* remove chat_box_tester
* remove some auto-generated files
* add **/ to .gitignore, because it seems just file name does not always work

Screenshot 2025-08-25 at 8 13 30 AM

### DIFF
--- a/pkgs/spikes/catalog_gallery/pubspec.yaml
+++ b/pkgs/spikes/catalog_gallery/pubspec.yaml
@@ -17,10 +17,10 @@ dependencies:
     path: ../../../examples/travel_app
 
 dev_dependencies:
+  dart_flutter_team_lints: ^3.5.2
   flutter_test:
     sdk: flutter
 
-  dart_flutter_team_lints: ^3.5.2
 
 flutter:
   uses-material-design: true


examples/travel_app/lib/travel_app.dart
gspencergoog, line 1: Don't you want to just remove this file, since it's generated?

pkgs/spikes/catalog_gallery/ios/Runner/Runner-Bridging-Header.h
gspencergoog, line 1: Same here (shouldn't this be deleted?)

pkgs/spikes/catalog_gallery/analysis_options.yaml
gspencergoog, line 1: Can you make this use the same options as the other analysis_options.yaml in the repo? That way we'll be consistent.

examples/travel_app/lib/travel_app.dart
polina-c: Nope: It is not generated.

I just created it to expose the catalog to the gallery app.

gspencergoog: Okay, if it's not generated, then why are you excluding it in the .gitignore?

gspencergoog: Oh!  Nevermind.  This is just GitHub getting the diff very wrong.  It was saying it got copied from `GeneratedPluginRegistrant.swift`.  But it has nothing to do with that.

pkgs/spikes/catalog_gallery/ios/Runner/Runner-Bridging-Header.h
gspencergoog: Nevermind.  Same diff issue with GitHub.

PR #188, opened 2025-08-25
Visualize example data for catalog items.

polina-c:
Contributes to https://github.com/flutter/genui/issues/131

Screenshot 2025-08-25 at 1 13 10 PM

### DIFF
--- a/examples/minimal_genui/android/settings.gradle.kts
+++ b/examples/minimal_genui/android/settings.gradle.kts
@@ -20,9 +20,6 @@ pluginManagement {
 plugins {
     id("dev.flutter.flutter-plugin-loader") version "1.0.0"
     id("com.android.application") version "8.9.1" apply false
-    // START: FlutterFire Configuration
-    id("com.google.gms.google-services") version("4.3.15") apply false
-    // END: FlutterFire Configuration
     id("org.jetbrains.kotlin.android") version "2.1.0" apply false
     // START: FlutterFire Configuration
     id("com.google.gms.google-services") version("4.3.15") apply false
--- a/examples/simple_chat/android/app/build.gradle.kts
+++ b/examples/simple_chat/android/app/build.gradle.kts
@@ -1,8 +1,5 @@
 plugins {
     id("com.android.application")
-    // START: FlutterFire Configuration
-    id("com.google.gms.google-services")
-    // END: FlutterFire Configuration
     id("kotlin-android")
     // The Flutter Gradle Plugin must be applied after the Android and Kotlin Gradle plugins.
     id("dev.flutter.flutter-gradle-plugin")
--- a/examples/minimal_genui/android/app/build.gradle.kts
+++ b/examples/minimal_genui/android/app/build.gradle.kts
@@ -1,8 +1,5 @@
 plugins {
     id("com.android.application")
-    // START: FlutterFire Configuration
-    id("com.google.gms.google-services")
-    // END: FlutterFire Configuration
     id("kotlin-android")
     // The Flutter Gradle Plugin must be applied after the Android and Kotlin Gradle plugins.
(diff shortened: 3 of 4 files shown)


examples/minimal_genui/android/settings.gradle.kts
```diff
@@ -20,6 +20,9 @@ pluginManagement {
 plugins {
     id("dev.flutter.flutter-plugin-loader") version "1.0.0"
     id("com.android.application") version "8.9.1" apply false
+    // START: FlutterFire Configuration
+    id("com.google.gms.google-services") version("4.3.15") apply false
```
gspencergoog: This looks bad.. Does the tool really duplicate entries that are already there, and already marked with its configuration comments?

examples/simple_chat/android/app/build.gradle.kts
```diff
@@ -1,5 +1,8 @@
 plugins {
     id("com.android.application")
+    // START: FlutterFire Configuration
+    id("com.google.gms.google-services")
```
gspencergoog: This also looks wrong...

pkgs/spikes/catalog_gallery/lib/main.dart
```diff
@@ -60,29 +60,55 @@ class _MainPageState extends State<MainPage> {
...
+      (CatalogItem item) => item.exampleData != null,
+    );
+
+    for (final item in items) {
+      final data = item.exampleData!;
+      final surfaceId = item.name;
+      _genUi.surfaceManager.addOrUpdateSurface(surfaceId, data);
+      surfaceIds.add(surfaceId);
+    }
+  }
```
gspencergoog, line 90: Don't you want to also have a `didUpdateWidget` here in order to detect catalog updates?

pkgs/spikes/catalog_gallery/macos/Runner.xcodeproj/project.pbxproj
gspencergoog, line 1: This should be ignored, right?

pkgs/spikes/catalog_gallery/lib/main.dart
polina-c: I would want if it would be production code. But it is a small spike, not intended for production and dynamic updates of widgets. So, will save on it.

pkgs/spikes/catalog_gallery/macos/Runner.xcodeproj/project.pbxproj
polina-c: I tried and no, it should not. Things are failing. See the comment:

https://github.com/flutter/genui/blob/25c6da2d0b99e904f6b50e0cf72dffea7e07b105/.gitignore#L2

examples/minimal_genui/android/settings.gradle.kts
polina-c: reverted for now

i do not know what to do with them going forward

we may want to file bug against flutter...

pkgs/spikes/catalog_gallery/macos/Runner.xcodeproj/project.pbxproj
gspencergoog: Yes, I remember the comment.  Okay, but are we going to have to check in a new version each time we modify code?

polina-c: It maybe not about code modification, but about refresh of something: maybe flutter version, maybe something else.

My hypotheses - not every time, but once a week, for example.

Yes, DevTools is acting this way - sometimes merging strange auto-updated files.

examples/minimal_genui/android/settings.gradle.kts
polina-c: When I worked in DevTools, we just were submitting them without reading :)

PR #195, opened 2025-08-26
Implement SurfaceController

jacobsimionato:


pkgs/flutter_genui/lib/src/core/surface_controller.dart
```diff
@@ -0,0 +1,23 @@
...
+import '../model/ui_models.dart';
+
+/// A callback for when a user interacts with a widget.
+typedef UiEventCallback = void Function(UiEvent event);
+
+class SurfaceController {
+  SurfaceController({required this.definitionNotifier, required this.catalog});
+
+  final ValueNotifier<UiDefinition?> definitionNotifier;
+  final Catalog catalog;
```
gspencergoog, line 17: Is the catalog meant to be per-surface?  That might be hard to communicate to the generating LLM: I don't think we want to have to communicate the entire catalog with each request.  Or is this just meant to be a reference to a common central catalog?  I wonder if there's a way we could enforce that there's only one catalog per `GenUIManager`?  Maybe have it owned by the `GenUIManager` instead, and pass one of those to the `SurfaceController` constructor?

examples/minimal_genui/lib/main.dart
```diff
@@ -81,7 +81,7 @@ class _MyHomePageState extends State<MyHomePage> {
   Future<void> _triggerInference() async {
     _chatController.setAiRequestSent();
     try {
-      final response = await _aiClient.generateText(
+      final response = await _aiClient.generateText(conversation: _conversation);
         List.of(_chatController.conversation.value),
```
gspencergoog, line 85: This looks like a syntax error?

Issue #192, opened 2025-08-25
Make composite UI submitting ALL values to AI


polina-c:
Prerequisite of https://github.com/flutter/genui/issues/131

PR #196, opened 2025-08-26
Fix missing values by keeping values in central place.

polina-c:
Fixes https://github.com/flutter/genui/issues/192

(I checked option with handling events that we discussed yesterday, however it seems to add significant complexity to user app, and it is different between travel_app and minimal_genui, as they use different widgets, that act differently, so i ended up implementing what seems to be most straightforward for me, to create short term fix, with TODO to revisit the design after discussion on go/gen-ui-surface-controller-proposal, and with plan to create facade to provide smoother UX)

# Before

before

# After

Screenshot 2025-08-26 at 8 34 47 AM

### DIFF
--- a/pkgs/flutter_genui/lib/src/core/genui_manager.dart
+++ b/pkgs/flutter_genui/lib/src/core/genui_manager.dart
@@ -60,12 +60,7 @@ class GenUiManager {
   final _surfaces = <String, ValueNotifier<UiDefinition?>>{};
   final _updates = StreamController<GenUiUpdate>.broadcast();
 
-  // TODO: revisit this implementation after discussion
-  // on go/gen-ui-surface-controller-proposal.
-  //
-  // TODO: consider to make it listenable by copying implementation of
-  // ValueListenable for lists from DevTools implementation.
-  final valueStore = ValueStore();
+  final valueStore = WidgetValueStore();
 
   Map<String, ValueNotifier<UiDefinition?>> get surfaces => _surfaces;
 
--- a/pkgs/flutter_genui/lib/src/facade/genui_surface.dart
+++ b/pkgs/flutter_genui/lib/src/facade/genui_surface.dart
@@ -129,7 +129,7 @@ class _GenUiSurfaceState extends State<GenUiSurface> {
       (String childId) => _buildWidget(definition, childId),
       _dispatchEvent,
       context,
-      widget.manager.valueStore,
+      widget.manager.valueStore.forSurface(widget.surfaceId),
     );
   }
 
--- a/pkgs/flutter_genui/lib/src/model/catalog.dart
+++ b/pkgs/flutter_genui/lib/src/model/catalog.dart
@@ -44,7 +44,7 @@ class Catalog {
     Widget Function(String id) buildChild,
     DispatchEventCallback dispatchEvent,
     BuildContext context,
-    ValueStore valueStore,
+    JsonMap valueStore,
   ) {
     final widgetType = (data['widget'] as JsonMap).keys.firstOrNull;
     final item = items.firstWhereOrNull((item) => item.name == widgetType);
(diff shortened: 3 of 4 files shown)


pkgs/flutter_genui/lib/src/catalog/core_widgets/elevated_button.dart
```diff
@@ -39,12 +39,13 @@ final elevatedButton = CatalogItem(
...
         required dispatchEvent,
         required context,
+        required values,
       }) {
         final buttonData = _ElevatedButtonData.fromMap(data as JsonMap);
         final child = buildChild(buttonData.child);
         return ElevatedButton(
           onPressed: () => dispatchEvent(
-            UiActionEvent(widgetId: id, eventType: 'onTap', value: null),
+            UiActionEvent(widgetId: id, eventType: 'onTap', value: values),
```
jacobsimionato, line 48: I wonder if it makes sense for the framework to add the values to the event before passing it to application code, so that developers don't need to be responsible for this, and to make sure it's always consistent?

pkgs/flutter_genui/lib/src/catalog/core_widgets/checkbox_group.dart
```diff
@@ -97,12 +97,14 @@ final checkboxGroup = CatalogItem(
...
+        required values,
       }) {
         final checkboxData = _CheckboxGroupData.fromMap(data as JsonMap);
         return _CheckboxGroup(
           initialValues: checkboxData.values,
           labels: checkboxData.labels,
           onChanged: (newValues) {
+            values[id] = newValues;
             dispatchEvent(
               UiChangeEvent(
```
jacobsimionato, line 109: I think we can remove UiChangeEvent if we are going to switch to the 'values' approach. In this case, we should ensure that the TravelApp also uses the values correctly.

pkgs/flutter_genui/lib/src/core/genui_manager.dart
```diff
@@ -59,6 +60,13 @@ class GenUiManager {
   final _surfaces = <String, ValueNotifier<UiDefinition?>>{};
   final _updates = StreamController<GenUiUpdate>.broadcast();
 
+  // TODO: revisit this implementation after discussion
+  // on go/gen-ui-surface-controller-proposal.
+  //
+  // TODO: consider to make it listenable by copying implementation of
+  // ValueListenable for lists from DevTools implementation.
+  final valueStore = ValueStore();
```
jacobsimionato: I think the ValueStore should be scoped to a particular surface, because the values from surface A aren't especially relevant to a submission button event in surface B.

Also, can we make it so the data is read-only here?

pkgs/flutter_genui/lib/src/catalog/core_widgets/checkbox_group.dart
polina-c: Yes, it makes sense, if we decide it is long term approach.

While it is short term fix, we need both.

polina-c:
> Hey I like this approach! It seems to be heading towards Greg's idea of widget state. But let's avoid having many parallel systems for the same thing, so maybe we can remove UiChangeEvent in favor of this system, then massage it towards widget state. Intuitively, I wonder if the state should live in the widget itself or in SurfaceController perhaps. Curious what @gspencergoog thinks too.

Actually GenUiManager de-facto acts as shared controller for all serfaces. As we want all surfaces to play well with each other, it may make sense for them to share controller.

pkgs/flutter_genui/lib/src/catalog/core_widgets/checkbox_group.dart
jacobsimionato: Okay! I think it might be quite simple to get the travel app to use the same event handling approach as "simple chat", so my preference would be to delete UiChangeEvent and UiEventManager as part of this PR, but if it seems like too much to do in one step, then feel free to delete later.

pkgs/flutter_genui/lib/src/core/genui_manager.dart
gspencergoog: Yes, I agree with Jacob: it should be scoped per surface.

pkgs/flutter_genui/lib/src/catalog/core_widgets/checkbox_group.dart
polina-c: Created separate issue: https://github.com/orgs/flutter/projects/222/views/1?pane=issue&itemId=126246710

PR #199, opened 2025-08-26
Make it easier to construct a `Catalog` with items added or removed

### LINKED ISSUE #152: Ability to easily add a new CatalogItem to the core catalog
Currently, developers create a Gen UI Manager like:

```dart
// Where is the catalog defined?
final manager = GenUiManager.chat(aiClient: GeminiAiClient(systemInstruction: _chatPrompt));
```

It uses a core catalog by default.

To add a widget to the core catalog, they need to do something like:

```dart
final catalog = Catalog([
  elevatedButtonCatalogItem,
  columnCatalogItem,
  text,
  checkboxGroup,
  radioGroup,
  textField,
  myCustomNewWidget,
]);
final manager = GenUiManager.chat(aiClient: GeminiAiClient(systemInstruction: _chatPrompt))
```

or else
```dart
final catalog = Catalog([
  ...coreCatalog.items, // Maybe awkward to need to grab the items from some preexisting catalog?
  myCustomNewWidget,
]);
final manager = GenUiManager.chat(aiClient: GeminiAiClient(systemInstruction: _chatPrompt))
```

Let's update the API such that:
1. It's obvious where the catalog is passed. No default, because the point is that developers need to add their own catalog items.
2. There is an obvious way to augment the core catalog rather than replace it.

andrewkolos:
## Description 

Resolves #152. This change reworks the API for constructing a `Catalog` into something more discoverable and intuitive for new users while also being flexible for those who want to provide their own custom widgets.

1. The existing constructor for `Catalog` is now private. A new public factory constructor, `Catalog.create`, can be used by users for the same purpose. We do this to increase the discoverability of `Catalog.core` (discussed later).
2. `GenUiManager` now requires a `Catalog` to be explicitly passed to it.
3. A new static member, `Catalog.core` is functionally equivalent to the `Catalog` that would be instantiated by `GenUiManager` if the user did not provide one. The doc comment of `Catalog.core` leaves a breadcrumb that users can follow to discover what is in the core catalog.

    Change 2 makes it easier for users to discover the concept of a `Catalog`. Change 3 makes it easier to discover and use the core catalog of widgets we have (i.e. rather than having to find some standalone list of core components—see the parent issue).

4. `Catalog` has new instance methods for modifying a catalog in an immutable manner.
    a. `copyWith` is used for adding items (e.g. `Catalog.core.copyWith([myCustomItem])`)
    b. `copyWithout`is used for removing items (e.g. `Catalog.core.copyWithout([CoreCatalogItems.radioGroup.name])`
5. Users that only want to use a few items from the core catalog can reference the newly introduced `CoreCatalogItems` class, which contains `static` members for each item (e.g. `Catalog.custom([CoreCatalogItems.text])`).

## Pre-launch Checklist

- [x] I signed the [CLA].

[changelog]: ../CHANGELOG.md

### DIFF
--- a/pkgs/flutter_genui/lib/src/core/core_catalog.dart
+++ b/pkgs/flutter_genui/lib/src/core/core_catalog.dart
@@ -11,6 +11,19 @@ import '../catalog/core_widgets/text.dart' as text_item;
 import '../catalog/core_widgets/text_field.dart' as text_field_item;
 import '../model/catalog_item.dart';
 
+// A list of all core catalog items.
+//
+// Meant for use by this library only.
+final List<CatalogItem> coreItems = [
+  CoreCatalogItems.checkboxGroup,
+  CoreCatalogItems.column,
+  CoreCatalogItems.elevatedButton,
+  CoreCatalogItems.image,
+  CoreCatalogItems.radioGroup,
+  CoreCatalogItems.text,
+  CoreCatalogItems.textField,
+];
+
 /// A collection of standard catalog items that can be used to build simple
 /// interactive UIs.
 class CoreCatalogItems {
@@ -36,15 +49,4 @@ class CoreCatalogItems {
 
   /// A widget that displays an image.
   static final CatalogItem image = image_item.image;
-
-  /// A list of all core catalog items.
-  static final List<CatalogItem> all = [
-    elevatedButton,
-    column,
-    text,
-    checkboxGroup,
-    radioGroup,
-    textField,
-    image,
-  ];
 }
--- a/pkgs/flutter_genui/lib/src/model/catalog.dart
+++ b/pkgs/flutter_genui/lib/src/model/catalog.dart
@@ -6,7 +6,7 @@ import 'package:collection/collection.dart';
 import 'package:dart_schema_builder/dart_schema_builder.dart';
 import 'package:flutter/material.dart';
 
-import '../core/standard_catalog.dart';
+import '../core/core_catalog.dart';
 import '../primitives/logging.dart';
 import '../primitives/simple_items.dart';
 import 'catalog_item.dart';
@@ -32,7 +32,7 @@ class Catalog {
   /// A catalog of simple items sufficient for building simple interactive UIs.
   ///
   /// For the full list of core widgets, see [CoreCatalogItems].
-  static final Catalog core = Catalog._(CoreCatalogItems.all);
+  static final Catalog core = Catalog._(coreItems);
 
   /// The list of [CatalogItem]s available in this catalog.
   final List<CatalogItem> items;
--- a/pkgs/flutter_genui/lib/flutter_genui.dart
+++ b/pkgs/flutter_genui/lib/flutter_genui.dart
@@ -11,7 +11,7 @@ export 'src/catalog/core_widgets/image.dart';
 export 'src/catalog/core_widgets/radio_group.dart';
 export 'src/catalog/core_widgets/text.dart';
 export 'src/catalog/core_widgets/text_field.dart';
-export 'src/core/core_catalog.dart';
+export 'src/core/core_catalog.dart' hide coreItems;
 export 'src/core/genui_manager.dart';
 export 'src/facade/genui_surface.dart';
 export 'src/facade/to_refactor/chat_widget.dart';


pkgs/flutter_genui/lib/src/model/catalog.dart
```diff
@@ -22,11 +23,49 @@ import 'ui_event_manager.dart';
...
+
+  /// Returns a new [Catalog] instance containg the items from this catalog
+  /// with the specified items removed.
+  Catalog copyWithout(List<String> itemNames) {
+    final updatedItems = items
+        .where((item) => !itemNames.contains(item.name))
+        .toList();
+    return Catalog._(updatedItems);
+  }
+
```
andrewkolos, line 66: The implementation of `copyWith` and `copyWithout` perform linear scans of string comparisons (another linear scanning operation). The performance of these methods will degrade exponentially with the number of items in the catalog. If we foresee some catalogs having thousands of items, we might want to implement some hashing here (and the bookkeeping to support that). However, given that I am not aware of this being a likely case (and in the interest of keeping things simple), I've left this as is.

pkgs/flutter_genui/lib/src/model/catalog.dart
```diff
@@ -21,12 +22,49 @@ import 'ui_event_manager.dart';
...
-  const Catalog(this.items);
+  factory Catalog.create(List<CatalogItem> items) {
+    return Catalog._(items);
+  }
+
+  /// A standard catalog of items sufficient for building simple interactive
+  /// UIs.
+  ///
+  /// For the full list of core widgets, see [CoreCatalogItems].
+  static final Catalog core = Catalog._(CoreCatalogItems.all);
```
andrewkolos: I choose the name `core` because I noticed existing code documentation (and discussion on #152) referring to our built-in catalog items as such. That being said, if `standard` or `builtIn` or `foundation` or whatnot is feels better, let me know.

pkgs/flutter_genui/lib/src/model/catalog.dart
```diff
@@ -21,12 +22,49 @@ import 'ui_event_manager.dart';
 /// 3. It dynamically generates a [Schema] that describes the structure of all
 ///    supported widgets, which can be provided to the AI model.
 class Catalog {
+  const Catalog._(this.items);
+
   /// Creates a new catalog with the given list of [items].
-  const Catalog(this.items);
+  factory Catalog.create(List<CatalogItem> items) {
```
andrewkolos, line 28: While `Catalog.custom` initially felt like a better name for this because it better hints at why one would use it over `Catalog.core`, it's not 100% clear that this library will even ship with a core catalog coupled directly into it (or we might decouple it out not long after).

* If this does happen, I feel like `Catalog.create` becomes the much better name (`Catalog.custom` is a surprising name for the only constructor of a class that users are meant to call directly).
* If we, however, end up keeping the idea of a `core` catalog baked into the library, then `Catalog.create` is still an okay name.

Weighing this all out, `Catalog.create` feels like the safer name to choose.

pkgs/flutter_genui/lib/src/core/core_catalog.dart
```diff
@@ -2,21 +2,49 @@
...
+  static final CatalogItem radioGroup = radio_group_item.radioGroup;
+
+  /// A material design text field.
+  static final CatalogItem textField = text_field_item.textField;
+
+  /// A widget that displays an image.
+  static final CatalogItem image = image_item.image;
+
+  /// A list of all standard catalog items.
+  static final List<CatalogItem> all = [
```
andrewkolos: Having an `all` method here has the drawback of creating two ways to create a core/standard catalog: `Catalog.core` and `Catalog.create(CoreCatalogItems)`, which runs counter to my idea of optimizing the getting-started experience by having a simpler API.

Still, I hesitate to remove this method because I feel it makes it easier to introduce a new catalog item but forget to update the core catalog with in. If we have this `all` member, then contributors adding a new item will likely see this `all` list and update it.

One alternative I'm considering here is whether we can move this list outside of the `CoreCatalogItems` class (but keep it in the same file). `Catalog.core` would reference this list, but we wouldn't expose/export this to users.

pkgs/flutter_genui/lib/src/model/catalog.dart
jacobsimionato: SGTM!

pkgs/flutter_genui/lib/src/model/catalog.dart
jacobsimionato: Agreed - these will likely run once at the start of the application with only a handful or 10s of items, so it doesn't have to be optimal.

pkgs/flutter_genui/lib/src/model/catalog.dart
```diff
@@ -21,12 +22,48 @@ import 'ui_event_manager.dart';
...
+    return Catalog._(items);
+  }
+
+  /// A catalog of simple items sufficient for building simple interactive UIs.
+  ///
+  /// For the full list of core widgets, see [CoreCatalogItems].
+  static final Catalog core = Catalog._(coreItems);
 
   /// The list of [CatalogItem]s available in this catalog.
   final List<CatalogItem> items;
```
jacobsimionato, line 38: Maybe we should make this an Iterable instead so it's actually immutable?

pkgs/flutter_genui/lib/src/core/core_catalog.dart
andrewkolos: Resolved in commit [remove CoreCatalogItems.all, replace with an internal list](https://github.com/flutter/genui/pull/199/commits/2c7c3078905bb21e67ee38b268e2093cceae3a97)

pkgs/flutter_genui/lib/src/core/genui_manager.dart
```diff
@@ -54,7 +53,7 @@ class SurfaceRemoved extends GenUiUpdate {
 }
 
 class GenUiManager {
-  GenUiManager({Catalog? catalog}) : catalog = catalog ?? coreCatalog;
+  GenUiManager({required this.catalog});
```
jacobsimionato, line 56: Love this - thanks!

pkgs/flutter_genui/lib/src/model/catalog.dart
andrewkolos: `CoreCatalogItems.all` is no longer a thing, but my original comment here still holds.

andrewkolos:
The [failing CI check](https://github.com/flutter/genui/actions/runs/17256098856/job/48968121165?pr=199) reveals a fundamental problem with having any sort of "default" factory defined in `Catalog` itself: it creates a tight coupling between the `core` and `model` layer.

While I still think this change would create the best out-of-box experience for this library, it's hard to ignore this problem. My current plan is to close this PR[^1], at least for now.

[^1]: We can still proceed with the idea of having `copyWith` and `copyWithout` methods on `Catalog`, if we still like those. We can also make the change to make `Catalog` into an immutable structure as well.

jacobsimionato:
Hey why don't you just add a `Catalog.core()` factory in as a class extension in a separate file in the "core" layer, which can call the `Catalog.create()` constructor in the "model" layer?

andrewkolos:
> Hey why don't you just add a `Catalog.core()` factory in as a class extension in a separate file in the "core" layer, which can call the `Catalog.create()` constructor in the "model" layer?

I actually considered this and discarded initially because `extension` doesn't support adding constructors, only methods. But, I think you are indeed onto something, I think `extension type` does support constructors, so I'll give it a try. Thanks!

jacobsimionato:
No, no I was wrong - I figured it would be possible to add static extension methods (if not constructors) that can be invoked on the underlying class object, but that's not true, so I don't see how we can add this syntactic sugar. Sorry for the noise! I think your CoreCatalogItems class is still quite nice though. Maybe we can add the similar static APIs to that, e.g.

```dart
CoreCatalogItems.catalog();
CoreCatalogItems.catalogWith(myAdditionalCatalogItems);
```

I think extension types don't make sense here because the point syntactically is we ideally don't want a wrapper class.

andrewkolos:
Superseded by https://github.com/flutter/genui/pull/241

PR #200, opened 2025-08-27
Simplify Gemini AI Client

jacobsimionato:
This pull request simplifies the AI client architecture by renaming `GeminiAiClient` to `FirebaseAiClient` and streamlining the `AiClient` interface to only include a `generateText` method. This change removes the previous `generateContent` method and its associated structured output capabilities, as well as the built-in retry mechanism. The update impacts core client logic, example applications, and documentation, aiming for a more focused and simplified AI interaction model.

### Highlights

* **Client Renaming**: The `GeminiAiClient` has been renamed to `FirebaseAiClient` to better reflect its integration with Firebase AI services.
* **API Simplification**: The `AiClient` interface has been significantly simplified by removing the `generateContent` method. The `generateText` method is now the sole method for AI interaction, returning a `String` directly.
* **Removal of Retry Mechanism**: The built-in retry logic with exponential backoff, previously present in `GeminiAiClient`, has been removed from `FirebaseAiClient`. This makes the client less resilient to transient network errors or API quota exceptions.
* **Examples and Documentation Updates**: All example applications and documentation have been updated to reflect the new `FirebaseAiClient` name and its revised API, including changes to method calls and parameter usage.
* **Type Safety Improvement**: While a previous bot comment highlighted a potential unsafe type cast related to `generateContent` returning `Object?`, the updated `AiClient` interface now ensures type safety by having `generateText` return `String`, thus eliminating the need for such casts.

### DIFF
--- a/examples/travel_app/lib/main.dart
+++ b/examples/travel_app/lib/main.dart
@@ -486,5 +486,12 @@ contain the other widgets.
 }
 ```
 
-When updating or showing UIs, **ALWAYS** use the addOrUpdateSurface tool to supply them. Prefer to collect and show information by creating a UI for it. When showing an itinerary, don't return it as text, use an ItineraryWithDetails widget.
+When updating or showing UIs, **ALWAYS** use the addOrUpdateSurface tool to
+supply them. Prefer to collect and show information by creating a UI for it.
+When showing an itinerary, don't return it as text, use an ItineraryWithDetails widget.
+
+# Final text output
+After calling tools to update the UI, you can optionally respond with a short
+message explaining what you've done, but be sure to avoid repetition with the
+content already displayed in the UI.
 ''';


examples/travel_app/lib/main.dart
```diff
@@ -156,32 +155,10 @@ class _TravelPlannerPageState extends State<TravelPlannerPage> {
...
-        _conversation,
-        S.object(
-          properties: {
-            'result': S.boolean(
-              description: 'Successfully generated a response UI.',
-            ),
-            'message': S.string(
-              description:
-                  'A message about what went wrong, or a message responding to '
-                  'the request. Take into account any UI that has been '
```
gspencergoog, line 169: I added this part because it has a tendency to reiterate the contents of the UI, especially in the case of itineraries: it would display the UI, and the write out Markdown that described the itinerary again. Now this directive will need to be supplied by the developer in their prompt.

pkgs/flutter_genui/lib/src/ai_client/ai_client.dart
```diff
@@ -32,21 +31,10 @@ abstract interface class AiClient {
   /// Switches the AI model to the given [model].
   void switchModel(AiModel model);
 
-  /// Generates content from the given [conversation] and returns an object of
-  /// type [T] that conforms to the given [outputSchema].
+  /// Generates content from the given [conversation] and returns a [String]
   ///
   /// The [additionalTools] are added to the list of tools available to the
   /// AI model.
-  Future<T?> generateContent<T extends Object>(
```
gspencergoog, line 40: I worry a little about removing the structured output capability.  It means that the developer is stuck with text output instead of having a choice.  as It is, they can get structured responses in addition to a UI, which seems like a common use case to me.

jacobsimionato: I think it's a reasonable use case, but here we're implementing a workaround (defining an outputTool) to Gemini's limitation of not being able to do structured output + tool calling which is adding significant complexity to our code that I think is unnecessary. This workaround should live in developer code, or perhaps Firebase AI Logic code if anything.

For greenfield developers trying to get up and running quickly, I think a simple GeminiAiClient with text output only will be enough to get them started.

For more advanced developers, they will implement their own tool calling loop and can do the workaround themselves if they need it. Presumable, any truly agentic app will already have some tools defined, and so they must only be using text-only output, or else have implemented a workaround similar to us.

examples/travel_app/lib/main.dart
jacobsimionato: Ahhh I see! Making the behavior of the example apps decent is important. But I wonder if we can do it without shipping the tool workaround in our SDK code?

Here, there is already ~20 extra lines of code here in the developer app implementing a workaround for this behavior, and many more in the core SDK, so maybe this complexity could instead be spent updating the system prompt for the travel app?

For the travel app specifically, we might also integrate at a lower level and include our own tool calling loop in developer code in which case we could implement this "outputTool" pattern there.

gspencergoog: I could have chosen to move this text to the travel app prompt, and called `generateText` here instead, so these lines aren't required, it's just how I chose to implement it, to keep the prompt simple.

I don't see the workaround in the core SDK as that bad: we already had to support tool calling, so there's only about 30 lines in the core SDK (to create the "final output" tool and handle it).

PR #214, opened 2025-08-28
Fix Seth's app simple_chat.

polina-c:
Fixes https://github.com/flutter/genui/issues/203
Fixes https://github.com/flutter/genui/issues/204

Screenshot 2025-08-27 at 7 46 50 PM

https://github.com/user-attachments/assets/3a41d8ac-9f97-48a1-b9c8-423d5d307bf7

### DIFF
--- a/examples/simple_chat/lib/main.dart
+++ b/examples/simple_chat/lib/main.dart
@@ -41,7 +41,7 @@ class _ChatScreenState extends State<ChatScreen> {
   // TODO: pass model from FirebaseAIService
   late final UiAgent _uiAgent = UiAgent(
     '''
-    You are a helpful assistant who chat with user,
+    You are a helpful assistant who chats with user,
     giving exactly one response for each user message.
     Your responses should contain acknowledgment
     of the user message.


packages/flutter_genui/lib/src/core/genui_manager.dart
```diff
@@ -54,26 +55,37 @@ class SurfaceRemoved extends GenUiUpdate {
...
+  final _userMessages = StreamController<UserMessage>.broadcast();
 
   @override
   final valueStore = WidgetValueStore();
 
   Map<String, ValueNotifier<UiDefinition?>> get surfaces => _surfaces;
 
   @override
-  Stream<GenUiUpdate> get updates => _updates.stream;
+  Stream<GenUiUpdate> get aiMessages => _aiMessages.stream;
```
jacobsimionato: How about `surfaceUpdates`? I think `aiMessage` implies that this is a response turn in a conversational UI, but instead this is a level more abstract than that, an update event which could create, update, delete a surface.

packages/flutter_genui/lib/src/core/genui_manager.dart
```diff
@@ -126,7 +139,15 @@ class GenUiManager implements SurfaceBuilder {
       genUiLogger.info('Deleting surface $surfaceId');
       final notifier = _surfaces.remove(surfaceId);
       notifier?.dispose();
-      _updates.add(SurfaceRemoved(surfaceId));
+      _aiMessages.add(SurfaceRemoved(surfaceId));
     }
   }
+
+  Stream<UserMessage> get userMessages => _userMessages.stream;
```
jacobsimionato: Move this above, next to aiMessages etc. Also, how about we call it `interactionEvents`? Again, I think userMessage might imply it's a text message written by the user and be confusing. I know the type is `UserMessage` right now, but I also think that in the future we should change it to a new data structure that just clearly represents an interaction event, to avoid confusion of using the same class to represent an interaction event and text message.

packages/flutter_genui/lib/src/core/genui_manager.dart
```diff
@@ -126,7 +139,15 @@ class GenUiManager implements SurfaceBuilder {
...
+
+  @override
+  void onSubmit(String surfaceId) {
```
jacobsimionato: How about `onInteractionEvent`? It's not necessarily a submit button, there are other patterns that are possible.

packages/flutter_genui/lib/src/core/genui_manager.dart
```diff
@@ -126,7 +139,15 @@ class GenUiManager implements SurfaceBuilder {
...
+    final value = valueStore.forSurface(surfaceId);
```
jacobsimionato: This is why SurfaceController makes things simple - we don't need to implement maps from SurfaceId to values everywhere. There is just one map to SurfaceController, then all the surface-scoped state can live in there :-D

packages/flutter_genui/lib/src/facade/ui_agent.dart
```diff
@@ -24,26 +24,98 @@ class UiAgent {
     String instruction, {
     Catalog? catalog,
     this.onSurfaceAdded,
-    this.onSurfaceRemoved,
+    this.onSurfaceDeleted,
+    this.okToUpdateSurfaces = false,
+    this.onWarning,
   }) : _genUiManager = GenUiManager(catalog: catalog) {
+    final technicalPrompt = _technicalPrompt(
+      okToUpdate: okToUpdateSurfaces,
```
jacobsimionato, line 32: Can we solve this problem using the approach in https://github.com/flutter/genui/pull/155, to actually just not provide tools to do disallowed actions? The advantages of this approach are:
- More reliable - we can guarantee that invalid actions won't occur as they don't exist
- Less cognitive load on the LLM. Right now we're saying to the LLM "here are 3 options, but please only choose the first 2, never the third".

jacobsimionato: If you remove this part, I'm happy to resync my PR and add it on top of yours.

packages/flutter_genui/lib/src/facade/ui_agent.dart
```diff
@@ -24,26 +24,98 @@ class UiAgent {
...
+    } else if (update is SurfaceRemoved) {
+      if (onSurfaceDeleted == null) {
+        onWarning?.call(
+          'AI attempted to remove a surface (${update.surfaceId}), '
+          'but onSurfaceDeleted handler is not set.',
+        );
+        return;
+      }
+      final message = AiUiMessage(definition: {}, surfaceId: update.surfaceId);
+      _addMessage(message);
```
jacobsimionato, line 100: This approach of only appending to the conversation doesn't work well for "update" and "delete" actions, because it creates a confusing situation for the LLM where you have a conversation that looks like:

- "Plan a trip to new york"
- Created new_york_trip_plan surface <content>
- User clicked "Search" with state {tripLength: 5 days}
- Updated new_york_trip_plan surface <content>

This is just a short example, but it creates an unnecessary long and confusing conversation because:
- Interaction events typically don't need to live in the conversation forever after they have been handled
- The full UI state after every "update" is pasted into the conversation history, and this clouds the LLMs thinking and makes it pay attention to and reference unexpected surfaces. Also, it adds unnecessary input tokens. Once the trip has been regenerated, the LLM can forget about the old trip now.

In the travel_app, we were able to make it work much better by making it so that we update the conversation for update/delete calls, and also skip over interaction events that have already been handled.

Would you be able to do that here? Or else, maybe can design ui_agent to only every add new surfaces in a conversational style, in which case the current context handling is fine? I actually think this limitation might be good at the facade layer.

examples/simple_chat/lib/main.dart
```diff
@@ -40,10 +40,16 @@ class _ChatScreenState extends State<ChatScreen> {
   final List<MessageController> _messages = [];
   // TODO: pass model from FirebaseAIService
   late final UiAgent _uiAgent = UiAgent(
-    'You are a helpful assistant.',
+    '''
+    You are a helpful assistant who chat with user,
```
gspencergoog:
```suggestion
    You are a helpful assistant who chats with user,
```

packages/flutter_genui/lib/src/core/genui_manager.dart
polina-c: interactionEvents may mean both ways - from user and from AI

named it userInput

is it better?

packages/flutter_genui/lib/src/core/genui_manager.dart
polina-c: onInteractionEvent may mean both ways: from user and from Ai

in software 'submit' is something that triggers the application to send user input somewhere, that is exactly what we mean here

for example, TextField in Flutter has event `onSubmitted`, but there is no button here, it is to handle user hitting 'Enter'

renamed to 'onSubmitted' to be aligned with flutter

packages/flutter_genui/lib/src/core/genui_manager.dart
polina-c: When we get things landed, lets compare pros and cons on the doc.

Here is my raw thoughts:

1. 'Controller' is normally something that we give to developer to control widget state. I do not think we need to give something to developer who is using GenUiSurface, to control the widget state. May be we will need it for future use cases, but I would wait till we face them.  

2. Controller is tightly coupled with the widget. What if app renders the widget for a surface, then stops rendering it and then renders again. Does it mean controller will be recreated? Does it mean the state will be lost? Does it mean developer will need to worry about preserving state? What if developer renders it twice (for example on two tabs, that both happen to be alive)? Will the state be duplicated, or we will want developer to worry about this?

3. If we have controller, all catalog widgets with input will have to be stateful. Do we want this limitation? 

"There is just one map to SurfaceController, then all the surface-scoped state can live in there" - I do not see where exactly the code will be reduced or simplified. Now we have JsonMap instead of SurfaceController, where all the surface-scoped state lives. Where is the gain?

Yes, we can create special class GenUiSurfaceState, to use it instead of JsonMap. However, I do not see gain in putting it to widget state. Widget state is not reliable place, because developer can rerender widgets the way that the state will be lost. So, we will need to put special effort to preserve this state in GenUiManager. And then we will face the question: why not to store it in GenUiManager in the beginning?
 
And there can be other considerations that I may forget now. So, let's continue on the doc :).

packages/flutter_genui/lib/src/facade/ui_agent.dart
polina-c: So, you want to control allowed actions on GenUiManager level, right?

Yes, i like approach. 

However, if I remove this part now, the app will stop working well, because AI will start updating/deleting surfaces and it will be very confusing.

How about merge this PR as is, to have the app in working state, merge your PR, and then accurately integrate UiAgent to take advantage of API you suggested in https://github.com/flutter/genui/pull/155 ?

packages/flutter_genui/lib/src/facade/ui_agent.dart
polina-c: Yes, let's make UiAgent as functional as possible, without increasing complexity of the API surface.

Is it ok to handle it in next PRs?

polina-c: If we know how to handle history in 'right' way, do we want to store conversation history in core classes, not in facade?

polina-c: > Would you be able to do that here? Or else, maybe can design ui_agent to only every add new surfaces in a conversational style, in which case the current context handling is fine? I actually think this limitation might be good at the facade layer.

I would suggest to have number of facades:

UiAgent - more or less generic, with natural limitations, resulted from simplification of core API
UiChatAgent - limited to append only wrapper for UiAgent (I am not sure it makes sense as i do not see it to add much simplification, but maybe I am wrong)
UiDashboardChatAgent - facade do handle chat plus pinned dashboard

PR #215, opened 2025-08-28
Fix subtle issues in the way context is managed in gemini_ai_client and elsewhere

jacobsimionato:
The main changes are:
- Make GeminiAiClient accept a *copy* of the conversation history rather than taking a reference to the array and modifying it, which is unexpected and unpredictable. GeminiAiClient still adds extra stuff to the conversation temporarily just to manage the function calling state within a specific multi-part tool-enabled inference.
- Update the conversion methods so that AiUiMessage is represented in the conversation for context purposes. Now, on the user side, the idea is to always have AiUiMessage in the conversation, and never the tool calls. Later, we'll clean up the data structures to make this more clear and limited to the developer.
- No-longer include the full UI definition in the FunctionResult that is returned from function calls, because it is already shown in the function parameters. This avoids the LLM needing to look at two copies of the same information in the conversation data structure.
- Fix a bug where the LLM was not actually propagating the updated conversation properly in subsequent tool loop calls, because the conversion to Content was outside the loop.
- A few superficial changes 

What this actually changes:
- Avoid multiple copies of Gen UI surface state appearing in conversation history
- Confusing mental model for developers

### DIFF
--- a/packages/flutter_genui/lib/src/facade/ui_agent.dart
+++ b/packages/flutter_genui/lib/src/facade/ui_agent.dart
@@ -203,5 +203,8 @@ $deleteInstruction
 
 When you are asking for information from the user, you should always include at least one submit button of some kind or another submitting element (like carousel) so that the user can indicate that they are done
 providing information.
+
+After you have modified the UI, be sure to use the provideFinalOutput to give
+control back to the user so they can respond.
 ''';
 }


packages/flutter_genui/lib/src/model/chat_message.dart
```diff
@@ -169,16 +169,19 @@ final class AiUiMessage extends ChatMessage {
...
   AiUiMessage({required this.definition, String? surfaceId})
     : uiKey = UniqueKey(),
+      parts = [TextPart(definition.asContextDescriptionText())],
       surfaceId =
           surfaceId ??
           ValueKey(DateTime.now().toIso8601String()).hashCode.toString();
 
   /// The JSON definition of the UI.
-  final JsonMap definition;
+  final UiDefinition definition;
```
gspencergoog, line 178: Oooh, good catch.  Maybe modify the doc comment? It's more than just JSON.

packages/flutter_genui/lib/src/model/chat_message.dart
```diff
@@ -169,16 +169,19 @@ final class AiUiMessage extends ChatMessage {
...
 
   /// A unique key for the UI widget.
   final Key uiKey;
 
   /// The unique ID for this UI surface.
   final String surfaceId;
+
+  final List<MessagePart> parts;
```
gspencergoog, line 186: Maybe add a doc comment for why these are here (what are they used for)?

Issue #235, opened 2025-08-29
Retire UiChangeEvent and UiEventManager


polina-c:
Follow up the decision: https://github.com/flutter/genui/pull/196#discussion_r2302722389

User interactions should be handled via GenUiHist, instead of GenUiSurface.onEvent.

PR #236, opened 2025-08-29
Rename uiAction to submit.

polina-c:
Contributes to https://github.com/flutter/genui/issues/235


polina-c:
> Why submit? I think "action" is actually more descriptive. Not all UI actions will be submits. It could be a "back" action, for instance, in a page-based app, or another navigation action.

Got it. Thanks.

Issue #134, opened 2025-08-14
Add a "checkout flow" to the agentic travel app to demonstrate we can complete a real task


jacobsimionato:
Currently, the user of the travel app can create itineraries, but there is no fixed schema for an itinerary, and no guaruntees any of the data is "real".

In a real agentic app, the LLM would help the user with a task that would result in actions taking place that can only be accomplished with correct data provided in the correct structure. E.g. an LLM can provide a beautiful flight booking experience, but if the flight information is hallucinated, then you can't actually call a booking API at the end and complete the task. Even if the information is slightly inaccurate, you might mislead the end user.

To prove developers can build a real agentic experience around Gen UI, we should demonstrate a flow like this in the example app.

Requirements:
- LLM should create a bookable itinerary using "real" fake data from data fetching tools
- When the user clicks "Checkout" they should be taken to a page that allows for booking that itinerary. This could be via a "CheckoutPage" CatalogItem, some navigation call baked into ItineraryResult, or some other tool call that the LLM makes.
- The API which launches the booking page should pass only the IDs associated with the items e.g. the flight IDs for the booking system, the hotel ids for the hotel etc, and the parameters to book e.g. dates etc
- The human-written booking page logic should look up and display the actual data for those bookable things and display them as a confirmation, with with a book button. It should fail if the Ids are incorrect.

See mocks from Ryan:
<img width="1096" height="1295" alt="Image" src="https://github.com/user-attachments/assets/5295d7ca-8c8b-41d7-83b9-0177c438153d" />

Dependencies:
- https://github.com/flutter/genui/issues/40, which probably depends on https://github.com/flutter/genui/issues/125

PR #237, opened 2025-08-29
Improve prompting for images.

polina-c:
Contributes to https://github.com/flutter/genui/issues/134

Otherwise AI comes up with fake images.

### DIFF
--- a/examples/travel_app/lib/main.dart
+++ b/examples/travel_app/lib/main.dart
@@ -405,7 +405,7 @@ following asset images:
 
 ${_imagesJson ?? ''}
 
-If there is no needed image in assets, use text description instead of images.
+If you can't find an appropriate image in the assets, use a text description instead.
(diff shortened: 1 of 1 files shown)


examples/travel_app/lib/main.dart
```diff
@@ -401,9 +401,13 @@ are not using that, you can use an `ElevatedButton`. Only use `OptionsFilterChip
 widgets inside of a `FilterChipGroup`.
 
 If you need to use any images, try to find the most relevant ones from the
-following asset images. Do not make up new image names, only use these:
+following asset images:
+
 ${_imagesJson ?? ''}
 
+If there is no needed image in assets, use text description instead of images.
```
gspencergoog:
```suggestion
If you can't find an appropriate image in the assets, use a text description instead.
```

PR #241, opened 2025-08-29
make it easier to construct a `Catalog` with items added or removed

### LINKED ISSUE #152: Ability to easily add a new CatalogItem to the core catalog

andrewkolos:
# Description

Resolves #152. 

* Replaces `coreCatalog` with `CoreCatalogItems`, a class that exposes static getters to each `CatalogItem` that used to be in `coreCatalog`. Also has an `asCatalog` static method, which returns a `Catalog` containing all core items.
* Marks the `catalog` parameter of the constructor `GenUiManager` as `required`.
* `Catalog.items` is now exposed as an `Iterable` rather than a `List`. This prevents confusion around whether the catalog can be modified while in-use.
* `Catalog` gets two convenience methods for adding/removing items from a catalog (in an immutable fashion): `copyWith(List<CatalogItem> items)` and  `copyWithout(Iterable<String>)`. 

## Pre-launch Checklist

- [x] I signed the [CLA].

[changelog]: ../CHANGELOG.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,8 +42,8 @@ agent-based user experiences.
 
 ```dart
 // Adding your widgets into the catalog.
-final catalog = Catalog([
-  ...coreCatalog.items,
+// Start with the built-in catalog and add your own custom widgets.
+final catalog = CoreCatalogItems.asCatalog().copyWith([
   myCustomNewWidget,
 ]);
 


sethladd:
Awesome! Please update the code snippet we have in the README, too! It would be awesome to show one small example there.

PR #254, opened 2025-08-30
Retire UiChangeEvent and UiEventManager

polina-c:
Fixes https://github.com/flutter/genui/issues/235

Also fixes some violations of [flutter style:](https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#naming-rules-for-typedefs-and-function-variables)
When naming callbacks, use FooCallback for the typedef, onFoo for the callback argument or property, and handleFoo for the method that is called. If Foo is a verb, prefer the present tense to the past tense (e.g. onTap instead of onTapped).

### DIFF
--- a/packages/flutter_genui/test/catalog/core_widgets_test.dart
+++ b/packages/flutter_genui/test/catalog/core_widgets_test.dart
@@ -2,8 +2,6 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-import 'dart:async';
-
 import 'package:flutter/material.dart';
 import 'package:flutter_genui/flutter_genui.dart';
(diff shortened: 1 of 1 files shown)


packages/flutter_genui/lib/src/model/ui_event_manager.dart
jacobsimionato, line 1: Can we delete this file and put these callback definitions somewhere else? Kind of weird having a file called "ui_event_manager.dart" when we have no class with that name.

polina-c:
> Can you make sure that the travel app does send events containing the user state (e.g. selected chip values) correctly with this change?

As this change touches many files, merging it now to avoid conflicts. 
Will re-test travel app right away.

PR #256, opened 2025-08-30
Rename catalog so that it is clear that it is just example

polina-c:
Fixes https://github.com/flutter/genui/issues/255

### DIFF
--- a/examples/travel_app/lib/main.dart
+++ b/examples/travel_app/lib/main.dart
@@ -97,7 +97,7 @@ class _TravelPlannerPageState extends State<TravelPlannerPage> {
   @override
   void initState() {
     super.initState();
-    _genUiManager = GenUiManager(catalog: exampleCatalog);
+    _genUiManager = GenUiManager(catalog: travelAppCatalog);
     _eventManager = UiEventManager(callback: _onUiEvents);
     _aiClient =
         widget.aiClient ??
--- a/packages/catalog_gallery/lib/main.dart
+++ b/packages/catalog_gallery/lib/main.dart
@@ -35,7 +35,7 @@ class MainPage extends StatefulWidget {
 class _MainPageState extends State<MainPage> {
   final Map<String, Catalog> catalogs = {
     'core catalog': coreCatalog,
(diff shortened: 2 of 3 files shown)


packages/catalog_gallery/lib/main.dart
```diff
@@ -35,7 +35,7 @@ class MainPage extends StatefulWidget {
 class _MainPageState extends State<MainPage> {
   final Map<String, Catalog> catalogs = {
     'core catalog': coreCatalog,
-    'travel app catalog': travel_app.catalog,
+    'travel app catalog': travel_app.exampleCatalog,
```
jacobsimionato: I actually didn't know that catalog_gallery depends on travel_app. My preference would be to build this preview functionality into the travel_app itself as a demo feature rather than having a separate app that depends on parts of the travel_app. I think that will make the previewing more discoverable, and it means that when Seth is demoing the travel app, he can quickly show people the basic components that are being used.

WDYT? This it orthogonal to this change, and not blocking at all.

examples/travel_app/lib/main.dart
```diff
@@ -97,7 +97,7 @@ class _TravelPlannerPageState extends State<TravelPlannerPage> {
   @override
   void initState() {
     super.initState();
-    _genUiManager = GenUiManager(catalog: catalog);
+    _genUiManager = GenUiManager(catalog: exampleCatalog);
```
jacobsimionato: How about travelCatalog? I think so far, the travel app code is written with the tone of a "real" app, so I think we should continue to avoid words like "example". But, I do agree with you that it's helpful to make it clear that this is a travel-specific catalog, not the built-in one etc.

packages/catalog_gallery/lib/main.dart
polina-c: I like it.

Arguments against it: one more app to manage

Arguments for it:
* it is cleaner
* When user copies the example app, they will get gallery, that is good

We can create widget CatalogGallery in flutter_genui and use it for both genui gallery and travel app gellery.

PR #258, opened 2025-08-31
Update copyright pubspec to be more permissive in SDK version to make it work with Dart beta

jacobsimionato:


jacobsimionato:
This is silly - I just needed to update my Dart and Flutter :-D

PR #277, opened 2025-09-04
Add troubleshooting tips.

polina-c:


tool/refresh_firebase_template.sh
```diff
@@ -29,7 +29,6 @@ TOOL_DIR=$(dirname "$0")
 PROJECT_ID="fluttergenui"
 
 EXAMPLES=(
-    "minimal_genui"
```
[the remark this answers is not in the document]
polina-c: thanks

packages/flutter_genui/USAGE.md
```diff
@@ -132,8 +132,20 @@ For a complete example, refer to [main.dart in the minimal_genui example](../../
...
+
+### Higher minimum macOS version is required
+
+If you are getting the error similar to one below, re-clone the repo, and then rerun your copy of `tool/refresh_firebase_template.sh`.
+
+```
+The pod "Firebase/CoreOnly" required by the plugin "firebase_app_check" requires a higher minimum macOS deployment version than the plugin's reported minimum version.
+To build, remove the plugin "firebase_app_check", or contact the plugin's developers for assistance.
+Error: Error running pod install
+```
```
andrewkolos: If I understand correctly, `tool/refresh/firebase_template.sh` only works for the projects in the `examples` app. If a user of the library encounters this issue while working on one of their own projects that utilizes `flutter_genui`, this instruction will be confusing. What do you think about linking/referencing the existing "Configure Firebase" section above instead?

jacobsimionato: I agree with this! I'm approving anyway because I think this PR is a strict improvement, but I like the idea!

polina-c: makes sense

thanks

PR #286, opened 2025-09-08
Update text chip to display label when no text is entered

jacobsimionato:
E.g. if there is a "Destination" chip, it should say "Destination" and then if the user clicks on it and enters "New York", the chip should now say "New York" instead.


jacobsimionato:
I have addressed the feedback. Please take another look.

jacobsimionato:
I have addressed the feedback. Please take another look.

jacobsimionato:
I have addressed the feedback. Please take another look.

PR #293, opened 2025-09-10
Split gallery.

polina-c:
Following up this request: https://github.com/flutter/genui/pull/256#discussion_r2312644867

Screenshot 2025-09-11 at 12 12 59 PM

### DIFF
--- a/examples/travel_app/lib/src/travel_planner_page.dart
+++ b/examples/travel_app/lib/src/travel_planner_page.dart
@@ -1,3 +1,7 @@
+// Copyright 2025 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
 import 'dart:async';
 
 import 'package:dart_schema_builder/dart_schema_builder.dart';


packages/flutter_genui/lib/src/facade/gallery_app.dart
jacobsimionato, line 1: How about this go in a separate folder like "development_utilities" seeing as it's a dev-specific thing and doesn't have much to do with ui_agent etc? There are a couple of other dev utilities I'm imaging too, e.g. a thing that can track and visualize requests, so I think it would be nice to start a folder.

packages/catalog_gallery/lib/main.dart
jacobsimionato, line 1: I think this is an app rather than a package. I actually think it should be distributed with the `flutter_genui` package eventually, so maybe we can put it at `packages/flutter_genui/examples/gallery`. WDYT? Or else I'm fine with it staying here for now.

packages/flutter_genui/lib/src/facade/gallery_app.dart
```diff
@@ -0,0 +1,81 @@
...
+import '../model/catalog_item.dart';
+
+class CatalogGalleryApp extends StatelessWidget {
+  const CatalogGalleryApp(this.catalog, {super.key});
+
+  final Catalog catalog;
+
+  @override
+  Widget build(BuildContext context) {
+    return MaterialApp(
```
jacobsimionato: How about making this a screen that can be embedded in an existing app? That's much more flexible, seeing as people can easily wrap it in a MaterialApp if they want.

examples/travel_app/catalog_gallery/lib/main.dart
jacobsimionato, line 1: Rather than having a separate app for this, why don't we fold it it into the travel_app? E.g. there is an overflow menu item like "Gallery" which pushes this screen as a route?

Advantages:
- Don't have to maintain all the boilerplate files associated with yet another app
- We can more easily test this, because it will be at our fingertips in the travel app
- We can deploy this as part of the travel app on Firebase and use it in demos to explain the component library

We could maybe rename CatalogGalleryApp to just CatalogGallery seeing as it's a screen that can be put inside another app if you want.

polina-c: I love it. Good idea. Will do.

packages/flutter_genui/lib/src/facade/gallery_app.dart
polina-c: yes!

also created separate package-level file for dev exports. does it make sense for you?

let me know if you want it differently

Issue #291, opened 2025-09-10
dart_schema_builder/lib/src/formats.dart has bad email regex


RandalSchwartz:
**Describe the bug**
`dart_schema_builder/lib/src/formats.dart` has an incorrect regex for email.

**To Reproduce**
View the source:
```
'email': (value) {
    return RegExp(
      r"^[a-zA-Z0-9.a-zA-Z0-9.!#$%&'*+-/=?^_`{|}~-]+@[a-zA-Z0-9]+\.[a-zA-Z]+",
    ).hasMatch(value);
  },
```

**Expected behavior**
No, please don't use this regex to validate email.  It's wrong.  The shortest regex to validate an RFC822 email address is about 2400 characters, and is given in https://www.ex-parrot.com/pdw/Mail-RFC822-Address.html.  Anything shorter is WRONG, unless you're using a grammar, as package:email_validator properly does.  Use that for Dart/Flutter please.

**Additional context**
Common mistake, because most people don't know the facts about email. :)

sethladd:
Let's remove any attempt at email validation from this package. Instead, allow the user to plug in their own email validation logic.

RandalSchwartz:
Just to reiterate, I'm happy with email_validator as doing the right thing, but strongly opposed to putting a junk "almost" regex and perpetuating the "this short regex can match email" myth.

PR #298, opened 2025-09-12
 fix(email): use safe regex pattern for email validation

loveucifer:
Replaced broken email regex in dart_schema_builder with a correct, practical pattern:

    r'^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$'

Also fixed test failure by importing formatValidators via relative path.

All tests pass locally.


RandalSchwartz:
This update would not address #291, so if I have a vote, I vote :-1:.

PR #299, opened 2025-09-12
Add example data and expose onsubmit in catalog view.

polina-c:


polina-c:
gemini /describe

packages/flutter_genui/USAGE.md
```diff
@@ -153,6 +153,19 @@ If something is unclear or missing, please [create an issue](https://github.com/
...
+### Configure logging
+
+To observe interaction between app and agent, enable logging:
+
+```dart
+final logger = configureGenUiLogging(level: Level.ALL);
+logger.onRecord.listen((record) {
+  print(
+    '${record.loggerName}: ${record.message}',
+  );
```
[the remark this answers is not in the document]
polina-c: i prefer conciser logging

examples/travel_app/lib/src/catalog/travel_carousel.dart
```diff
@@ -70,6 +70,89 @@ final travelCarousel = CatalogItem(
           dispatchEvent: dispatchEvent,
         );
       },
+  exampleData: {
```
gspencergoog, line 73: This is a good example, but I wonder if it could be smaller.  The whole example will be included in the LLM context, so the smaller the better.  Of course, it still needs to be representative.

polina-c: On one hand, yes.

On another hand it is real example from real LLM response, that looks very good in catalog gallery and creates good test coverage.

Maybe in future we will want to provide more than one example per catalog, to show both simplest cases and advanced cases.

PR #305, opened 2025-09-14
Hwrd code fit to 'contain'.

polina-c:
Contributes to: https://github.com/flutter/genui/issues/134

Because:

1. We do not want images to be cut.
2. We do not want LLM to waste its attention to this thing with big risk to make it wrong.

### DIFF
--- a/examples/travel_app/integration_test/app_test.dart
+++ b/examples/travel_app/integration_test/app_test.dart
@@ -92,17 +92,13 @@ const Map<String, Object> _baliResponse = {
           {
             'id': 'bali_memorial_image',
             'widget': {
-              'Image': {
-                'fit': 'cover',
-                'assetName': 'assets/travel_images/bali_memorial.jpg',
-              },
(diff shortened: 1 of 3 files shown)


jacobsimionato (changes requested):
Hey I like the idea of hardcoding this for now, but I actually vote we use "cover" as the universal option.

I think the main goal of our built-in widget catalog right now is to help us make compelling demos, because in production use cases, developers can always make their own image widget which displays differently. In our demos, the images are used to add some flavor to the UI, but not as diagrams etc where the entire image needs to be visible. We have demo images which are not all the same aspect ration, but all roughly 3:4 or 4:3 etc. We have use cases which are varied - 1:1 ratio carousels and diagrams, header images etc. If we use "contain", then we will never get consistently shaped images in carousels etc which looks cluttered and unpolished. So let's use 'cover' so that in our layouts, we can have images of any size we like, without needing to have such a huge range of demo images.

PR #318, opened 2025-09-17
Hotel details.

polina-c:
Fixes https://github.com/flutter/genui/issues/309

Screenshot 2025-09-17 at 3 38 42 PM


examples/travel_app/lib/src/catalog/travel_carousel.dart
```diff
@@ -17,7 +20,13 @@ final _schema = S.object(
       description: 'A list of items to display in the carousel.',
       items: S.object(
         properties: {
-          'title': S.string(description: 'The title of the carousel item.'),
+          'description': S.string(
```
jacobsimionato, line 23: Optional: How about having a separate subtitle field or something which is shown in a different color, like a grey or something? While it's a small thing, I think it makes the UI look quite a bit more polished to have visually separated sections like that.

jacobsimionato (approved):

examples/travel_app/lib/src/catalog/travel_carousel.dart
polina-c: I am afraid it may make it more complicated for AI. So, we may loose reliability.
I like the idea. But I would make it P2 and test it carefully. BTW, can be a good item for external contributors.

polina-c:
Merged in spite of the error as it seems main is broken: https://github.com/flutter/genui/pull/322

Issue #319, opened 2025-09-17
Fix issues with images.


polina-c:
1. Make sure images are not cut
2. Update catalog to render only assets because we never use network and AI is getting confused. In real app the catalog item should detect if it is network image or asset image.

PR #325, opened 2025-09-18
Improve handling of images.

polina-c:
Fixes: https://github.com/flutter/genui/issues/319


examples/travel_app/lib/src/travel_planner_page.dart
```diff
@@ -435,7 +435,7 @@ ${_imagesJson ?? ''}
 list that might be tangentially relevant. DO NOT USE ANY IMAGES NOT IN THE LIST.
 It is fine if the image is irrelevant, as long as it is from the list.
 
-- Use assetName for images from the list only - NEVER use `url` and reference
+- Use location for images from the list only - NEVER use `url` and reference
```
jacobsimionato: Maybe update this prompting as necessary?

packages/flutter_genui/lib/src/catalog/core_widgets/image.dart
```diff
@@ -10,16 +10,7 @@ import '../../primitives/simple_items.dart';
...
-      description:
-          'The URL of the image to display. Only one '
-          'of url or assetName may be specified.',
-    ),
-    'assetName': S.string(
-      description:
-          'The name of the asset to display. Only '
-          'one of assetName or url may be specified.',
-    ),
+    'location': S.string(description: 'Asset path or network URL.'),
```
jacobsimionato: Describe the format in detail here - how can we tell apart the path or network URL? This is what the LLM will use to understand this field properly.

PR #336, opened 2025-09-24
Adds a .guides folder with content

### LINKED ISSUE #334: Create a .guides folder and make agent guidelines discoverable there
Now that the usage guide for `flutter_genui` is finished, we should add a `.guides` folder and maintain agent-facing guidance there.

redbrogdon:
Adds a `.guides` folder for `flutter_genui` with agent-facing guidance and an example. Also fixes a type in `USAGE.md`.

There are some open questions here:
* Is this the right guidance for agents?
* How can the example be tested in CI?
* How can we avoid repeated content between README.me, USAGE.md, and these new files?

Fixes #334.


packages/flutter_genui/.guides/config.json
```diff
@@ -0,0 +1,19 @@
...
+      "name": "Connecting to an Agent Provider",
+      "description": "Instructions for connecting the app to an agent provider like Firebase AI Logic.",
+      "path": "docs/connect_to_agent_provider.md"
+    },
+    {
+      "name": "Creating a custom CatalogItem",
+      "description": "Instructions for creating your own widgets and adding them to an agent's catalog.",
+      "path": "docs/create_a_custom_catalogitem.md"
+    }
+  ]
```
mbleigh: You don't need any of this actually - just having files in `docs` will get them detected and examples are not quite ready for primetime yet (you could move that to a `docs/*.md` for now if you want it surfaced and just make the description that it's an example).

For what you're doing here no `config.json` should be necessary.

packages/flutter_genui/.guides/setup.md
```diff
@@ -0,0 +1,72 @@
...
+      flutter_genui:
+        git:
+          url: https://github.com/flutter/genui.git
+          path: packages/flutter_genui
+          ref: 6e472cf0f7416c31a1de6af9a0d1b4cc37188989
+      flutter_genui_firebase_ai:
+        git:
+          url: https://github.com/flutter/genui.git
+          path: packages/flutter_genui_firebase_ai
+    ```
```
mbleigh: If they're running `/setup` they will have already installed the `flutter_genui` package -- that's where the guidance is being loaded from. So you should assume that they are already in a Flutter app with the package installed.

redbrogdon: Got it. Removed.

packages/flutter_genui/.guides/config.json
redbrogdon: Awesome. One less file is always nice.

redbrogdon:
> Overall this looks great! Have you actually tried this out with Gemini CLI to see how it works?

Yes. It's clearly getting context to Gemini CLI, which is great.

I'm not sure how to help Gemini stitch together the guidance, though. When I hand it just the `USAGE.md` file that's in the package directory, it seems easy for Gemini to hand me a plan for setting up and using the package that matches what I've got described in the guide. With the same guidance spread out among different files in `.guides`, though, it doesn't have as easy a time putting that holistic plan together.

sethladd (approved):
RSLGTM so we can test it

redbrogdon:
Pardon the force-pushes to my fork. Just me trying to remember how git works.

PR #342, opened 2025-09-30
Remove dev-only root library.

polina-c:
See flutter/flutter practice: https://api.flutter.dev/flutter/rendering/Layer/debugDisposed.html

Follow up for internal discussion: https://chat.google.com/room/AAQAL-XDdiM/5dXrSe5GL2Y/5dXrSe5GL2Y?cls=10


jacobsimionato (approved):
Thanks!!

PR #359, opened 2025-10-08
Make the package web compatible.

gspencergoog:
# Description

This makes the package use a conditional import for the only place it imports `dart:io`, so that the package can be web compatible.

This pull request makes the `json_schema_builder` package web-compatible by refactoring its `SchemaCache` to use conditional imports for platform-specific implementations of file system access. It introduces an abstract `SchemaCacheBase` for common logic and separate concrete implementations for web and non-web environments. The changes also improve error handling by throwing specific exceptions instead of returning null, integrate logging, and ensure proper resource management for HTTP clients. This refactoring directly addresses several critical and high-priority feedback points from previous reviews, significantly improving maintainability and robustness.

### Highlights

* **Intent**: This pull request aims to make the `json_schema_builder` package web-compatible. It achieves this by refactoring the `SchemaCache` class to use conditional imports for platform-specific implementations, specifically to handle the `dart:io` dependency which is not available on web platforms. This also addresses prior feedback regarding code duplication and resource management.
* **Changes**: The core changes involve a significant refactoring of the schema caching mechanism:
1.  **Conditional Imports:** The original `schema_cache.dart` now uses conditional exports to load either `schema_cache_file.dart` (for non-web platforms, using `dart:io`) or `schema_cache_web.dart` (for web platforms, which throws an `UnimplementedError` for file schemes).
2.  **Abstract Base Class:** A new `SchemaCacheBase` abstract class (`schema_cache_base.dart`) was introduced. It encapsulates common caching, HTTP fetching, and JSON parsing logic, and defines an abstract `getCachedFile` method for platform-specific file access.
3.  **Error Handling:** A new `SchemaFetchException` class (`exceptions.dart`) was added. The `SchemaCacheBase.get` method now throws this exception instead of returning `null` on errors (e.g., unsupported schemes, failed HTTP requests, file access issues). `SchemaRegistry` and `SchemaValidation` were updated to handle these new exceptions.
4.  **Resource Management:** A `close()` method was added to `SchemaCacheBase` to properly dispose of internally created `http.Client` instances. `SchemaRegistry` now includes a `dispose()` method that calls `_schemaCache.close()`.
5.  **Logging Integration:** `LoggingContext` was integrated into `SchemaCacheBase`, `SchemaRegistry`, and `ValidationContext` to provide better debugging information.
6.  **Dependency Update:** The `pubspec.yaml` was updated to add `meta` as a dependency, likely for annotations like `@protected`.
* **Reviewer Activity**: This pull request directly addresses several high-priority and medium-priority concerns raised by `gemini-code-assist[bot]` in previous reviews:
*   **Code Duplication:** The refactoring into `SchemaCacheBase` and platform-specific implementations resolves the high-priority concern about code duplication between `SchemaCache` implementations.
*   **Unnecessary Constructor on Interface:** The refactoring implicitly resolved the medium-priority issue of an unnecessary constructor on an interface by moving common logic to an abstract base class.
*   **Silent Exception Swallowing:** The change from returning `null` to throwing `SchemaFetchException` and integrating `LoggingContext` addresses the high-priority concern about silently swallowing exceptions.
*   **Synchronous I/O in Async Method:** The `getCachedFile` implementation in `schema_cache_file.dart` was corrected to use `readAsString()` (asynchronous I/O) and the redundant `uri.scheme == 'file'` check was removed, resolving a critical concern.
*   **Unclosed HTTP Client:** The introduction of `close()` in `SchemaCacheBase` and `dispose()` in `SchemaRegistry` addresses the medium-priority concern about unclosed `http.Client` instances.

<details>
<summary><b>Changelog</b></summary>

* **added**
    * packages/json_schema_builder/lib/src/exceptions.dart
    * packages/json_schema_builder/lib/src/schema_cache_base.dart
    * packages/json_schema_builder/lib/src/schema_cache_file.dart
    * packages/json_schema_builder/lib/src/schema_cache_web.dart
* **changed**
    * packages/json_schema_builder/lib/src/schema_cache.dart
    * packages/json_schema_builder/lib/src/schema_registry.dart
    * packages/json_schema_builder/pubspec.yaml
    * packages/json_schema_builder/lib/src/schema_validation.dart
    * packages/json_schema_builder/test/schema_cache_test.dart
    * packages/json_schema_builder/test/test_suite_test.dart
* **removed**
</details>


gspencergoog:
> Awesome! I guess this works fine on web so far because it's dart:io, so if we don't use schema caching then we never hit the exceptions.

Right.  I noticed this though, because the pub site checks for dart:io on the web and won't mark the package as web compatible.

> Another way of structuring the conditional imported part could be to use more composition and less inheritance, e.g. have a purely abstract CachedSchemaFileLoader class with two implementations rather than a larger partially abstract one. This looks great though!

You're right, that's better.  Done!

Issue #358, opened 2025-10-08
Use centralized constant declarations for JSON schema keys, rather than literals


jacobsimionato:
Based on feedback on https://github.com/flutter/genui/pull/354 from @polina-c :

We have a lot of magic string literals in our JSON schema declarations and decoding logic e.g. "widget", "id" etc.

Let's centralize these as constants somewhere to make them less prone to breaking through typos in maintenance.

@polina-c :

There are two parts in this question:
1. Private constants
2. Public: constants that are used by the client applications

polina-c:
On one hand, I agree. On other hand:

1. It will make the code longer, and thus harder to read
2. _P is a enum-like class and enums are expected to have all elements to be of the same kind, and to have 'values' as a special field.

So, there are pros and cons.

@jacobsimionato , can you make a call here?

gspencergoog:
... Or, you can call it `_values`, since it's a private class anyhow.  Then you get the best of both.  Although it still doesn't address problems with fields that are called by Dart keywords.

jacobsimionato:
Sorry I missed this before!

I think I agree with the Greg's underlying point that we should make it super clear that these constants are field identifiers, not the values themselves, and avoid symbol conflicts. Given these are all going to be inside a parent class, I don't think they should technically conflict with an A2UI-related local variable e.g. `value` because  this would be accessed like `_P.value` right?

But, I still like the idea of making it super explicit that these are field names.

Also, do we need to implement `values` here? AFAICT these are not so much like an enum, and more just a list of constants. Do we ever need to iterate over the values, or do an exhaustive match against them, or define a typed variable which must be one of these values?

How about:
```dart
final class _Fields {
  static const mainAxisAlignment = 'mainAxisAlignment';
  static const crossAxisAlignment = 'crossAxisAlignment';
  static const children = 'children';
  static const spacing = 'spacing';
  static const classField = 'class';
  static const values = 'values';
}
```

jacobsimionato:
I'm honestly fine with any of these options though. This may be controversial, but I'd probably lean towards leaving things as they are with magic literals in here. I think that can make the code easier to read and verify, because there is less indirection.

PR #374, opened 2025-10-10
Clean up hard coded literals.

polina-c:
Contributes to https://github.com/flutter/genui/issues/358

If looks good, will ask AI to update other catalog items the same way.

### DIFF
--- a/packages/flutter_genui/lib/src/primitives/simple_items.dart
+++ b/packages/flutter_genui/lib/src/primitives/simple_items.dart
@@ -7,7 +7,7 @@ typedef JsonMap = Map<String, Object?>;
 T parseEnum<T>(String? value, List<T> values, T defaultValue) {
   if (value == null) return defaultValue;
   return values.firstWhere(
-    (T e) => e.toString().split('.').last == value,
+    (T e) => (e as Enum).name == value,
     orElse: () => defaultValue,
   );
 }


packages/flutter_genui/lib/src/catalog/core_widgets/column.dart
```diff
@@ -47,52 +42,33 @@ extension type _ColumnData.fromMap(JsonMap _json) {
...
+    _P.crossAxisAlignment.name: crossAxisAlignment,
   });
 
   List<String> get children =>
-      ((_json['children'] as List?) ?? []).cast<String>();
-  double get spacing => (_json['spacing'] as num?)?.toDouble() ?? 8.0;
-  String? get mainAxisAlignment => _json['mainAxisAlignment'] as String?;
-  String? get crossAxisAlignment => _json['crossAxisAlignment'] as String?;
+      ((_json[_P.children.name] as List?) ?? []).cast<String>();
+  double get spacing => (_json[_P.spacing.name] as num?)?.toDouble() ?? 8.0;
```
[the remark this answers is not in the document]
polina-c: 1. It is ok to have two separate values in this case. 
2. I want to limit this PR to a concrete small clean up.

packages/flutter_genui/lib/src/primitives/simple_items.dart
```diff
@@ -3,3 +3,11 @@
 // found in the LICENSE file.
 
 typedef JsonMap = Map<String, Object?>;
+
+T parseEnum<T>(String? value, List<T> values, T defaultValue) {
```
jacobsimionato, line 7: This is great!

Nit: Can we have named params just so it's a bit more obvious what the parameters mean at the call sites?

packages/flutter_genui/lib/src/catalog/core_widgets/column.dart
```diff
@@ -9,32 +9,27 @@ import 'package:json_schema_builder/json_schema_builder.dart';
 import '../../model/catalog_item.dart';
 import '../../primitives/simple_items.dart';
 
+enum _P { mainAxisAlignment, crossAxisAlignment, children, spacing }
```
jacobsimionato, line 12: I think maybe constant strings are a better fit here than an enum, seeing as the constants here aren't really interchangeable instances of something, i.e. there are no specific polymorphism requirements.

E.g. maybe `mainAxisAlignmentPropertyName` and `crossAxisAlignmentPropertyName`.

packages/flutter_genui/lib/src/catalog/core_widgets/column.dart
jacobsimionato, line 1: I'd love to see what @gspencergoog things too before we go and change the entire codebase. I think I'm strongly in favor of the parseEnum utility, and mildly in favor of using the constants.

packages/flutter_genui/lib/src/primitives/simple_items.dart
gspencergoog: I agree with Jacob about the named parameters.

Also, it's too bad we can't eliminate the `List<T> values` and just use `T.values` in the implementation.  (I know we can't, it's just too bad).

packages/flutter_genui/lib/src/catalog/core_widgets/column.dart
gspencergoog: So, how would that look?  Something like this?:

```dart
abstract base class FauxEnum {
	String get name;
	List<String> get values;
}

final class _P extends FauxEnum {
  const _P(this.name);
  String name;
  static const mainAxisAlignment = _P('mainAxisAlignment');
  static const crossAxisAlignment = _P('crossAxisAlignment');
  static const children = _P('children');
  static const spacing = _P('spacing');
  List<String> get values => [
    mainAxisAlignment,
    crossAxisAlignment,
    children,
    spacing,
  ].map<String>((v) => v.name).toList();
}
```

And I guess you'd have to make a base class for the class `FauxEnum` with the static strings that has the value accessor in order for the parseEnum function to work similarly.  But maybe you could eliminate the values arg to parseEnum then, since `values` wouldn't have to be static?

This is a nice cleanup, but one issue I have with all of this (including the enums) is that if there is a property that ends up being a Dart keyword, then we could have issues. (e.g. `static const String class = _P('class');` wouldn't work).  The current implementation doesn't have that problem.

At least with the constant strings, you could still implement it, it's just that the name would be different from the value: `static const String clazz = _P('class');`

polina-c:
Thanks for the comments. I will respond and then I will make sure we all have agreement here before merging (it is good that it is not urgent). So, it will wait for Jacob to be back from OOO.

packages/flutter_genui/lib/src/catalog/core_widgets/column.dart
jacobsimionato: Oh I propose that we keep using the enums for MainAxisAlignment, CrossAxisAlignment here. These are from Flutter and they make sense as enums, and the parseEnum function is neat and works well in my opinion. These enums aren't used to define the string "MainAxisAlignment", but rather the different alignments that are available.

For the property names e.g. "mainAxisAlignment", "crossAxisAlignment", I vote we instead just have something like:

```dart
const mainAxisAlignmentPropertyName = `mainAxisAlignment`;

final _schema = S.object(
  properties: {
    mainAxisAlignmentPropertyName: S.string(
      description:
          'How children are aligned on the main axis. '
          'See Flutter\'s MainAxisAlignment for values.',
      enumValues: MainAxisAlignment.values.map((e) => e.name).toList(),
```

packages/flutter_genui/lib/src/primitives/simple_items.dart
jacobsimionato: Thanks for iterating. I'd still like to see the named parameters here before submitting this!

PR #380, opened 2025-10-13
Remove lockfiles and add to gitignore

gspencergoog:
# Description

Remove `pubspec.lock` files from source control and add them to `.gitignore` for package directories.


packages/json_schema_builder/.gitignore
```diff
@@ -0,0 +1 @@
+pubspec.lock
```
polina-c, line 1: Why not add it to the root gitignore?

tool/test_and_fix/.gitignore
```diff
@@ -1,3 +1,4 @@
 # https://dart.dev/guides/libraries/private-files
 # Created by `dart pub`
 .dart_tool/
+pubspec.lock
```
polina-c: and this one

packages/json_schema_builder/.gitignore
gspencergoog: Because we don't want to ignore it at the root level: we want to check in `pubspec.lock` files in the "application" packages, but not in the "library" packages.  But actually, I just looked, and they are already supposed to be ignored everywhere. I guess these two were already checked in for some reason? In any case, we only want to ignore them in the library-like packages (which means I need to look again at what `test_and_fix` is doing: it's an "application", so it should keep its pubspec.lock.)

I'll update this PR.

polina-c: Can you remind me why we want to check them in for applications and what is the process to upgrade them?

gspencergoog: See https://dart.dev/tools/pub/private-files#pubspec-lock

PR #381, opened 2025-10-13
Add height limit per item for gallery view.

polina-c:
Contributes to https://github.com/googlecloud-appeco-incubator/gemini-flow/issues/31

Also removes deprecated providers.


polina-c:
the logic turned out to be more complicated

so, i am not merging this and creating custom catalog view in the app

PR #410, opened 2025-10-21
Refactor the custom backend to work with A2UI-based schema

jacobsimionato:
This PR refactors the `custom_backend` example to fully align with the A2UI streaming protocol specification.

Key changes include:

- **Protocol Compliance:** The app no longer uses a custom `uiGenerator` tool. It now correctly generates a `surfaceUpdate` tool call that conforms to the official `A2uiSchemas.surfaceUpdateSchema`.
- **A2UI Message Parsing:** The response from the LLM is now parsed into a strongly-typed `A2uiMessage` object using the `A2uiMessage.fromJson` factory.
- **`BeginRendering` Implementation:** A `BeginRendering` message is now synthetically created and sent to the `GenUiManager` after the `SurfaceUpdate` message, correctly following the A2UI progressive rendering flow.
- **Constant `surfaceId`:** A constant `kSurfaceId` has been introduced to ensure consistency across the prompt, UI, and message creation.
- **Testing:** The tests have been updated to validate the new A2UI-compliant message flow. The saved test responses have been updated with real data captured from the Gemini API.
- **Cleanup:** Removed the unused `widget_test.dart` file.

### DIFF
--- a/examples/custom_backend/lib/protocol/protocol.dart
+++ b/examples/custom_backend/lib/protocol/protocol.dart
@@ -7,7 +7,8 @@ import 'package:flutter_genui/flutter_genui.dart';
 import '../backend/api.dart';
 import '../backend/model.dart';
 import '../debug_utils.dart';
-import '../main.dart';
+
+const kSurfaceId = 'custom_backend_surface';
 
 class Protocol {
   Future<List<A2uiMessage>?> sendRequest(


examples/custom_backend/lib/main.dart
```diff
@@ -12,6 +12,7 @@ void main() {
 }
 
 const _title = 'Custom Backend Demo';
+const kSurfaceId = 'custom_backend_surface';
```
polina-c: Using prefix 'k' is debatable, that means not everyone will follow this code style.

https://github.com/flutter/flutter/issues/20858

this tradition is very old and came from C:

"c" was the tag for type "char", so it couldn't also be used for "const"; so "k" was chosen, since that's the first letter of "konstant" in German, and is widely used for constants in mathematics

https://stackoverflow.com/questions/5016622/where-does-the-k-prefix-for-constants-come-from

How about declaring genui traditions for debatable things in some central Coding Style document in the repo, to make it clear for all developers?

PR #31, opened 2025-07-31
Update GEMINI instructions for catalog items, add implementation descriptions.

gspencergoog:
# Description

Just added a hint to use extension types for the data in catalog items, and added some files that outline the implementation of the travel app and the SDK package.

PR #435, opened 2025-10-23
Factor out integration code from example to library, so that client apps can use it and it is test covered here.

polina-c:
Contributes to #31

### DIFF
--- a/examples/custom_backend/test/backend_api_test.dart
+++ b/examples/custom_backend/test/backend_api_test.dart
@@ -20,7 +20,7 @@ void main() {
     }
 
     // To update the saved responses, run the app, select "Request Gemini",
-    // and copy the console output of the "Response body" to the
+    // and copy the console output of the "full-response" to the
     // corresponding `saved-response-X.json` file in `assets/data/`.
     test(
       'sendRequest works for $savedResponse',


examples/custom_backend/lib/protocol.dart
```diff
@@ -0,0 +1,41 @@
...
+class Protocol {
+  final Catalog catalog = CoreCatalogItems.asCatalog();
+
+  Future<ParsedToolCall?> sendRequest(
+    String request, {
+    required String? savedResponse,
+  }) async {
+    final techPrompt = genUiTechPrompt([_toolName]);
+
+    final toolCall = await Backend.sendRequest(
```
jacobsimionato: Shouldn't the backend be initiating these tool calls? Shouldn't it be something like:
1. On startup, flutter sends tool declarations and system prompt fragment to the backend
2. Backend does inferences whenever it wants and sends Gen UI messages back to Flutter

examples/simple_chat/firebase-debug.log
jacobsimionato, line 1: Let's remove this and .gitignore it

packages/flutter_genui/lib/src/facade/direct_call_integration/utils.dart
```diff
@@ -0,0 +1,73 @@
...
+
+/// Prompt to be provided to the LLM about how to use the UI generation tools.
+String genUiTechPrompt(List<String> toolNames) {
+  final toolDescription = toolNames.length > 1
+      ? 'the following UI generation tools: '
+            '${toolNames.map((name) => '"$name"').join(', ')}'
+      : 'the UI generation tool "${toolNames.first}"';
+
+  return '''
+To show generated UI to user, use $toolDescription.
```
jacobsimionato, line 20: Keep in mind that this is being used with a confidential product, so this content is fine, but we won't be able to put product-specific stuff in here e.g. references to the specific catalog items etc.

jacobsimionato (approved):

examples/simple_chat/firebase-debug.log
polina-c: done

packages/flutter_genui/lib/src/facade/direct_call_integration/utils.dart
polina-c: Sure!

The goal here is to construct a prompt that can be used in any app.

examples/custom_backend/lib/protocol.dart
polina-c: yes, it will be closer to what we want

polina-c: updated, thank you for the great suggestion!

PR #438, opened 2025-10-23
Refactors the GenUI API to create `ContentGenerator`

gspencergoog:
# Description

This pull request introduces a significant refactoring of the GenUI API, moving away from a direct `AiClient` dependency to a more abstract `ContentGenerator` interface. This change aims to enhance modularity, remove the need for client-side LLM connections, and improve the overall architecture for handling generative UI conversations. Key updates include a new `DataPath` system for efficient data model management and increased flexibility in core UI components, alongside comprehensive updates to example applications and tests to align with the new API.

Key improvements include:
- Decoupling `flutter_genui` from a specific client implementation.
- Greatly simplifying `GenUiConversation` and application-level widgets like `TravelPlannerPage` by encapsulating AI interaction logic within the `ContentGenerator`.
- Introducing a typed `DataPath` class, which makes data model interactions more robust.
- Adding support for template-based lists, rows, and columns.

### Highlights

* **API Refactoring**: The core GenUI API has been refactored to replace the `AiClient` interface with a new `ContentGenerator` interface, decoupling the framework from specific AI client implementations and removing the dependency on client-side LLM connections.
* **Data Model Enhancements**: A new `DataPath` class has been introduced for more robust and efficient path management within the `DataModel`, improving how data is accessed and updated. The `_notifySubscribers` method in `DataModel` was also optimized for better performance.
* **GenUiConversation Updates**: The `GenUiConversation` class now orchestrates interactions using the new `ContentGenerator` and `GenUiManager`, streamlining the conversation flow and handling UI updates, text responses, and errors through dedicated streams.
* **UI Component Flexibility**: The `List` and `Row` core UI components have been made more flexible. The `ListView.builder` now uses `shrinkWrap: true` and `ClampingScrollPhysics` for better layout, and the `Row` widget no longer implicitly wraps children in `Expanded`, allowing for more granular layout control by the AI.
* **Firebase AI Integration**: The `flutter_genui_firebase_ai` package has been updated to implement the new `ContentGenerator` interface, with the `FirebaseAiClient` being renamed to `FirebaseAiContentGenerator`.

<details>
<summary><b>Changelog</b></summary>

* **examples/simple_chat/lib/main.dart**
    * Replaced `FirebaseAiClient` with `FirebaseAiContentGenerator` for content generation.
    * Updated `GenUiConversation` to use the new `contentGenerator` parameter.
    * Changed `onWarning` callback to `onError` for more structured error handling.
* **examples/travel_app/integration_test/app_test.dart**
[...]

PR #439, opened 2025-10-23
This creates the `flutter_genui_a2ui` package, which implements `A2uiContentGenerator`

gspencergoog:


gspencergoog:
[](@jacobsimionato The only files different from https://github.com/flutter/genui/pull/438 are the ones in packages/flutter_genui_a2ui, so you can ignore files outside of that.

The CI checks are going to fail until I can merge the other one and reduce the file count.

In reality, there are only a few files you need to review here (the rest are boilerplate):

```
M       packages/flutter_genui/lib/src/model/data_model.dart
A       packages/flutter_genui_a2ui/CHANGELOG.md
A       packages/flutter_genui_a2ui/DESIGN.md
A       packages/flutter_genui_a2ui/GEMINI.md
A       packages/flutter_genui_a2ui/IMPLEMENTATION.md
A       packages/flutter_genui_a2ui/LICENSE
A       packages/flutter_genui_a2ui/README.md
A       packages/flutter_genui_a2ui/analysis_options.yaml
A       packages/flutter_genui_a2ui/example/README.md
A       packages/flutter_genui_a2ui/example/analysis_options.yaml
A       packages/flutter_genui_a2ui/example/lib/main.dart
A       packages/flutter_genui_a2ui/example/pubspec.yaml
A       packages/flutter_genui_a2ui/lib/flutter_genui_a2ui.dart
A       packages/flutter_genui_a2ui/lib/src/a2ui_agent_connector.dart
A       packages/flutter_genui_a2ui/lib/src/a2ui_content_generator.dart
A       packages/flutter_genui_a2ui/pubspec.yaml
A       packages/flutter_genui_a2ui/test/a2ui_agent_connector_test.dart
A       packages/flutter_genui_a2ui/test/a2ui_content_generator_test.dart
A       packages/flutter_genui_a2ui/test/fakes.dart
```

packages/flutter_genui_a2ui/example/lib/main.dart
```diff
@@ -0,0 +1,173 @@
...
+  late final GenUiConversation _uiAgent;
+
+  @override
+  void initState() {
+    super.initState();
+    _contentGenerator = A2uiContentGenerator(
+      serverUrl: Uri.parse('http://localhost:10002'),
+    );
+    _uiAgent = GenUiConversation(
+      contentGenerator: _contentGenerator,
```
jacobsimionato, line 55: Yes! This is the dream! So simple!!

packages/flutter_genui_a2ui/lib/src/a2ui_agent_connector.dart
```diff
@@ -0,0 +1,234 @@
...
+  /// A stream of errors from the A2A connection.
+  Stream<Object> get errorStream => _errorController.stream;
+
+  /// Fetches the agent card.
+  ///
+  /// The agent card contains metadata about the agent, such as its name,
+  /// description, and version.
+  Future<AgentCard> getAgentCard() async {
+    final card = await client.getAgentCard();
+    return AgentCard(
```
jacobsimionato, line 72: Good idea abstracting this so that we can swap out the A2A library later

packages/flutter_genui_a2ui/lib/src/a2ui_agent_connector.dart
```diff
@@ -0,0 +1,234 @@
...
+  final Uri url;
+
+  final _controller = StreamController<genui.A2uiMessage>.broadcast();
+  final _errorController = StreamController<Object>.broadcast();
+  @visibleForTesting
+  late A2AClient client;
+  @visibleForTesting
+  String? taskId;
+  @visibleForTesting
+  String? contextId;
```
jacobsimionato: I wonder if contextId should actually be a constructor parameter so that people can load previous conversations etc

packages/flutter_genui_a2ui/lib/src/a2ui_agent_connector.dart
```diff
@@ -0,0 +1,234 @@
...
+    if (data.containsKey('surfaceUpdate') ||
+        data.containsKey('dataModelUpdate') ||
+        data.containsKey('beginRendering') ||
+        data.containsKey('deleteSurface')) {
+      if (!_controller.isClosed) {
+        _log.finest(
+          'Adding message to stream: '
+          '${jsonEncode(data)}',
+        );
+        _controller.add(genui.A2uiMessage.fromJson(data));
```
jacobsimionato, line 222: 🎉 It works!

packages/flutter_genui_a2ui/lib/src/a2ui_content_generator.dart
```diff
@@ -0,0 +1,76 @@
...
+        _textResponseController.add(responseText);
+        _addMessage(AiTextMessage([TextPart(responseText)]));
+      }
+    } finally {
+      _isProcessing.value = false;
+    }
+  }
+
+  void _addMessage(ChatMessage message) {
+    _conversation.value = [..._conversation.value, message];
```
jacobsimionato: Given that conversation isn't actually used here at all, seeing as the server is stateful, how about removing it and letting that be managed within the flutter_genui or application layer?

This seems to be an observable conversation which includes *just* the text prompts and text responses and not the generated UIs, so I don't see how it's useful. It's not suitable for rendering because it doesn't have the text and UI interleaved as is conventional. Also, it's duplicative of the _textResponseController which already exposes this information, so let's just rely on that.

jacobsimionato (approved):
I think we can delete `packages/flutter_genui_a2ui/android|ios|web|macos|linux` etc right?

That should just be a library with no platform directories, and then we should have platform directories for the example, e.g. `packages/flutter_genui_a2ui/example/android`

packages/flutter_genui_a2ui/lib/src/a2ui_content_generator.dart
gspencergoog: Yeah, agreed.

packages/flutter_genui_a2ui/lib/src/a2ui_agent_connector.dart
gspencergoog: Yeah, let's go with that.  Seems useful.

gspencergoog:
I'm going to commit this because it "works", and it updates the API for `ContentGenerator.sendRequest` to a much better abstraction, but there are still some issues with the data model not working right with indices in lists.

Issue #465, opened 2025-10-28
It appears that genui repo doesn't resolve dependencies with flutter stable


sethladd:
I have the latest flutter stable (3.35.7), and when I try to run `flutter pub get`, I get:

```
sethladd@Seths-MacBook-Air genui % flutter pub get
Resolving dependencies... 
Note: meta is pinned to version 1.16.0 by flutter_test from the flutter SDK.
See https://dart.dev/go/sdk-version-pinning for details.


Because flutter_genui depends on flutter_test from sdk which depends on meta 1.16.0, meta 1.16.0 is required.
So, because flutter_genui_a2ui depends on a2a ^3.0.0 which depends on meta ^1.17.0, version solving failed.
```

I assume this means that our repo doesn't work with flutter stable?

yjbanov:
The offending line seems to be in the `a2a` package: https://github.com/shamblett/a2a/blob/09b5a748c116bc7d1937c26bb18e117e2e88c6a7/pubspec.yaml#L28

sethladd:
I'm running into other dependency problems trying to use flutter stable. I think we should just land the PR that communicates that we run on beta. And fix up dependencies and versions to also work on stable as another PR

PR #466, opened 2025-10-28
explicitly require flutter > 3.37, and minor cleanup in README

sethladd:


sethladd:
fixes https://github.com/flutter/genui/issues/465

gspencergoog:
I have a PR out to fix this on the a2a package (see https://github.com/shamblett/a2a/pull/52).  I think we should hold off on this change until we see if they are responsive and we get the change in there. If they take another day or two, then we can just land this and limit things to beta. I'd like to be able to support Flutter stable, though.

sethladd:
well, right now we don't support flutter stable :)

we can land this now, and if a2a uses an older meta, then we can update our pubspec

sethladd:
maybe there's a way to do an override of meta, on our end?

sethladd:
I'd like to prefer that someone can clone the repo and have the tools tell them what version of Flutter they need. As we expand to support more versions of Flutter, we can broaden the Flutter SDK constraints.

I do agree that the repo should work with flutter stable and flutter beta, but while we're moving fast, it's OK if we work only with flutter beta

gspencergoog:
Okay, sure.  We can try and do an override, and then we can turn on stable testing again too.  If that doesn't work, we can limit to beta.

jacobsimionato:
Override sounds good to me. Otherwise, I'm in favor of submitting this PR to at least make our compatibility requirements explicit.

sethladd:
ok, let me try to put in an override

sethladd:
as mentioned, I'm having other version dependency issues trying to get back to stable. For now, I think we practically work on beta, so we should land this. Then, later we can make all our deps work on stable.

jacobsimionato:
SGTM, let's do it

PR #468, opened 2025-10-29
Genai content generator

sethladd:


packages/flutter_genui_google_generative_ai/lib/src/google_content_converter.dart
```diff
@@ -0,0 +1,138 @@
...
+            google_ai.Part(
+              functionResponse: google_ai.FunctionResponse(
+                id: part.callId,
+                name: part.callId,
+                // The result from ToolResultPart is a JSON string
+                response: protobuf.Struct.fromJson(
+                  jsonDecode(part.result) as Map<String, Object?>,
+                ),
+              ),
+            ),
```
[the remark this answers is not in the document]
sethladd: this might be valid

sethladd:
Sonnet says:

The Problem

The ToolResultPart class only has two fields:
callId - The ID of the call
result - The result data
However, Google's FunctionResponse requires three fields:
id - The call ID
name - The function name
response - The result data
In google_content_converter.dart lines 98-110 and 119-137, the converter sets name: part.callId, which is incorrect. The name field should be the function name, not the call ID.

Root Cause

This is an API design limitation in the core flutter_genui package. The ToolResultPart model was designed for Firebase AI, which only requires a call ID (Firebase's FunctionResponse constructor takes just callId and response). Google's API has stricter requirements.

When This Bug Manifests

This bug occurs when:
A user passes chat history containing ToolResponseMessage objects to sendRequest()

The converter must translate ToolResultPart → google_ai.FunctionResponse

Without the function name, the API receives incorrect data

Note: The bug is currently latent because the content generators work with provider-specific types internally and never create ToolResponseMessage objects. However, it would break if users manually construct history with tool results.

Recommended Fix

Update the core model in /packages/flutter_genui/lib/src/model/chat_message.dart:
```
final class ToolResultPart implements MessagePart {
  final String callId;
  final String toolName; // ADD THIS FIELD
  final String result;
   const ToolResultPart({    required this.callId,    required this.toolName,  // ADD THIS PARAMETER    required this.result,  });
}
```

Then update the converters to use part.toolName for the name field.
This is a breaking change that requires coordinated updates across all packages, but it's the correct architectural fix to support multiple AI providers properly

sethladd:
@jacobsimionato @gspencergoog what do you think about the advice above, from Sonnet? Note that this PR is trying to use the new generative language package, for a simpler developer experience just trying out GenUI.

examples/simple_chat/README.md
```diff
@@ -35,11 +36,45 @@ The application's logic is contained almost entirely within `lib/main.dart`.
 
 ## Getting Started
 
-To run this application, you will need to have a Firebase project set up and configured.
+This example supports two AI backends: **Google Generative AI** (default) and **Firebase AI**. Choose the option that best fits your needs.
 
-1. **Configure Firebase**: Follow the instructions in the main `flutter_genui` [USAGE.md](../../packages/flutter_genui/USAGE.md) to add Firebase to your Flutter app. You will need to replace the placeholder values in `lib/firebase_options.dart` with the configuration from your own Firebase project.
-2. **Run the App**: Once Firebase is configured, you can run the app like any other Flutter project:
+### Option 1: Using Google Generative AI (Default)
```
jacobsimionato, line 41: Put a warning here that this is for demo purposes, and Firebase AI should be used for productionization.

examples/simple_chat/lib/main.dart
```diff
@@ -8,13 +8,38 @@ import 'package:flutter/material.dart';
...
+import 'firebase_options_stub.dart';
 import 'package:logging/logging.dart';
 
+/// Enum for selecting which AI backend to use.
+enum AiBackend {
+  /// Use Firebase AI
+  firebase,
+
+  /// Use Google Generative AI
+  google,
```
jacobsimionato: Make this name more descriptive. How about googleGenerativeAi ?

packages/flutter_genui_google_generative_ai/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+If neither is provided, the package will attempt to use the default environment variable.
+
+## Differences from Firebase AI
+
+This package uses the Google Cloud Generative Language API instead of Firebase AI Logic. Key differences:
+
+- **Server-side**: This API is meant for Dart command-line, cloud, and server applications
+- **Flutter apps should use Firebase AI**: For mobile and web applications, consider using `flutter_genui_firebase_ai` instead, which provides client-side access
+
```
jacobsimionato, line 53: Can we omit this example below and point to the main README? Seeing as this is an ancillary package, it'd be great to avoid duplicating this content, as it creates more places to keep in sync.

packages/flutter_genui_google_generative_ai/lib/src/google_generative_service_interface.dart
```diff
@@ -0,0 +1,48 @@
...
+/// This class is used to wrap the Google Cloud AI [google_ai.GenerativeService]
+/// so that it can be used interchangeably with other implementations of the
+/// [GoogleGenerativeServiceInterface].
+class GoogleGenerativeServiceWrapper
+    implements GoogleGenerativeServiceInterface {
+  /// Creates a new [GoogleGenerativeServiceWrapper] that wraps the given
+  /// [_service].
+  GoogleGenerativeServiceWrapper(this._service);
+
+  final google_ai.GenerativeService _service;
```
jacobsimionato, line 35: Does this object have configurable options which it remembers? If it doesn't, then we can get rid of this level of indirection.

sethladd:
Chatted with @jacobsimionato , he advised to just set the `name:` to a blank string, as this code will likely be removed in the future. We left a comment for why we set that parameter to an empty string.

packages/flutter_genui_google_generative_ai/lib/src/google_content_converter.dart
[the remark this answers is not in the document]
sethladd: we discussed this, and I learned that this code path will be deleted later anyway. so we set this to an empty string, and left a comment why

examples/simple_chat/README.md
sethladd: Good idea

examples/simple_chat/lib/main.dart
sethladd: yeah, much better

packages/flutter_genui_google_generative_ai/lib/src/google_generative_service_interface.dart
sethladd: this pattern was taken from the Firebase AI Logic implementation. See gemini_generative_model.dart from the AI Logic impl.

packages/flutter_genui_google_generative_ai/README.md
sethladd: good idea, done

PR #474, opened 2025-10-30
Implement flex weight for components.

gspencergoog:
# Description

This implements the `weight` attribute for components as an `Expanded` with a flex factor for components that have `weight` set.

### DIFF
--- a/packages/flutter_genui/lib/src/catalog/core_widgets/row.dart
+++ b/packages/flutter_genui/lib/src/catalog/core_widgets/row.dart
@@ -133,11 +133,11 @@ final row = CatalogItem(
                   mainAxisSize: MainAxisSize.min,
                   children: childIds
                       .map(
-                        (id) => buildWeightedChild(
-                          componentId: id,
+                        (componentId) => buildWeightedChild(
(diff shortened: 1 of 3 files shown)


packages/flutter_genui/lib/src/catalog/core_widgets/row.dart
```diff
@@ -110,19 +110,34 @@ final row = CatalogItem(
...
+                  ),
+                  crossAxisAlignment: _parseCrossAxisAlignment(
+                    rowData.alignment,
+                  ),
+                  children: childIds.map((id) {
+                    final component = getComponent(id);
+                    final weight = component?.weight;
+                    final childWidget = buildChild(id, dataContext);
+                    if (weight != null) {
+                      return Expanded(flex: weight, child: childWidget);
```
jacobsimionato: Neat! Previously, I had assumed that introducing these interactions between child/parent widgets would add too much complexity to the rendering engine to make sense, but I guess it's not actually that bad. Now I see how we could implement the "hero image" pattern that Paul is after for full bleed images at the top of cards.

PR #473, opened 2025-10-30
Implement primary button, make data model more permissive.

gspencergoog:
# Description

This adds support for the primary button tag, as well as several other updates to match the latest schema.

## Summary of Changes

This pull request significantly enhances the UI generation framework by introducing support for a distinct "primary" button type and making the underlying data model more robust and permissive. It also includes comprehensive updates to align with the latest schema definitions, refactors data subscription mechanisms for greater flexibility, and improves dynamic list rendering. These changes collectively aim to provide a more expressive and debuggable system for building generative UIs.

### Highlights

* **Primary Button Support**: Introduced a 'primary' boolean property to the Button component, allowing it to be styled distinctly for important actions and updated example data to demonstrate its usage.
* **Data Model Permissiveness and Parsing**: Enhanced the data model's update method to more flexibly handle root path updates, accepting both lists and maps, and improved parsing of complex data structures, including correct handling of list indices in data paths.
* **Schema Updates**: Synchronized the `server_to_client.json` schema with the latest changes, including new `weight` and `Icon` properties, and clarified data path descriptions to use '/' as a separator.
* **Unified Data Subscription**: Refactored data subscription methods from `subscribeToStringArray` to `subscribeToObjectArray` to support a broader range of object types, improving flexibility in data binding.
* **Improved List Rendering**: Updated `List` and `ComponentChildrenBuilder` to handle template-based lists using maps of components, offering more flexible dynamic UI generation.
* **Enhanced Logging**: Added more detailed logging across various core components for better debugging and insight into UI and data model operations, including formatted JSON output for payloads.

<details>
<summary><b>Changelog</b></summary>

* **examples/travel_app/lib/src/catalog/checkbox_filter_chips_input.dart**
    * Updated `dataContext.subscribeToStringArray` to `dataContext.subscribeToObjectArray` for broader type support.
* **examples/travel_app/lib/src/travel_planner_page.dart**
    * Added an `onSurfaceUpdated` callback to `GenUiConversation` to trigger `_scrollToBottom()` when the UI surface is updated.
* **packages/flutter_genui/lib/src/catalog/core_widgets/button.dart**
    * Added a `primary` boolean property to the `Button` schema and its data model (`_ButtonData`).
    * Implemented styling for primary buttons, applying `colorScheme.primary` for background and `colorScheme.onPrimary` for foreground when `primary` is true.
    * Updated example data to demonstrate both primary and secondary buttons.
* **packages/flutter_genui/lib/src/catalog/core_widgets/image.dart**
[...]

PR #476, opened 2025-10-30
Add Icon Component

gspencergoog:
# Description

Adds support for the `Icon` component. Only added a limited set of icons, to keep the description short, and to keep from overwhelming the context with them.  There are 8825 different icons, so it would be a lot...

This PR currently also contains the contents of the https://github.com/flutter/genui/pull/473 PR, which will be rebased away once that lands.

### DIFF
--- a/packages/flutter_genui/lib/src/catalog/core_widgets/icon.dart
+++ b/packages/flutter_genui/lib/src/catalog/core_widgets/icon.dart
@@ -1,0 +1,147 @@
+// Copyright 2025 The Flutter Authors.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:flutter/material.dart';
+import 'package:json_schema_builder/json_schema_builder.dart';
+
+import '../../model/a2ui_schemas.dart';
+import '../../model/catalog_item.dart';
+import '../../model/data_model.dart';
+import '../../primitives/simple_items.dart';
+
+final _schema = S.object(
+  properties: {
+    'name': A2uiSchemas.stringReference(
+      description:
+          '''The name of the icon to display. This can be a literal string ('literalString') or a reference to a value in the data model ('path', e.g. '/icon/name').''',
+      enumValues: AvailableIcons.allAvailable,
+    ),
+  },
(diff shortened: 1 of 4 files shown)


packages/flutter_genui/lib/src/catalog/core_widgets/icon.dart
```diff
@@ -0,0 +1,125 @@
...
+      _IconData.fromMap({'name': name});
+
+  JsonMap get nameMap => _json['name'] as JsonMap;
+
+  String? get literalName => nameMap['literalString'] as String?;
+  String? get namePath => nameMap['path'] as String?;
+}
+
+const Map<String, IconData> _iconMap = {
+  'add': Icons.add,
```
jacobsimionato: Have we agree on this set of icons with Paul etc?

packages/flutter_genui/lib/src/catalog/core_widgets/icon.dart
```diff
@@ -0,0 +1,125 @@
...
+import 'package:json_schema_builder/json_schema_builder.dart';
+
+import '../../model/a2ui_schemas.dart';
+import '../../model/catalog_item.dart';
+import '../../model/data_model.dart';
+import '../../primitives/simple_items.dart';
+
+final _schema = S.object(
+  properties: {
+    'name': A2uiSchemas.stringReference(
```
jacobsimionato, line 15: Why a string here rather than an enum? Wouldn't an enum use the same number of input and output tokens and provide validation?

packages/flutter_genui/lib/src/catalog/core_widgets/icon.dart
gspencergoog: No, I just picked some.  These are OK for a starting point, but we surely need to curate this a bit.

packages/flutter_genui/lib/src/catalog/core_widgets/icon.dart
gspencergoog: Good Idea, I'll try that. I can make the description smaller then, which is good.

PR #480, opened 2025-11-01
add copyright to jules_setup

sethladd:


sethladd:
this will make some of our CI bots happy. weird that this wasn't caught before?

PR #482, opened 2025-11-02
Provide surfaceId to widgetBuilder.

polina-c:
Fixes https://github.com/flutter/genui/issues/481

Issue #481, opened 2025-11-02
Provide surfaceId to widget builder.


polina-c:
@jacobsimionato , there should be a way to get surface id in widget builder, because the builder needs to load the remembered state on re-render.

```
final questionGroup = CatalogItem(
  name: 'QuestionGroup',
  dataSchema: _schema,
  widgetBuilder:
      ({
        required data,
        required id,
        required buildChild,
        required dispatchEvent,
        required context,
        required dataContext,
      }) { ???
```

Yes, I could handle data processing as you explained in https://github.com/flutter/genui/issues/446.
 (see PR 162)

However, there are concerns.

While for some applications it may be better to use the recommended way,

in case of my app the code complexity would reduce significantly, if surfaceId was provided. Without surfaceId the code is distributed between classes, takes a lot of lines and is not easy to read and understand.

Also, I am scared that something will stop working inside genui and I will not see why.

The specifics of my app are:

1. 'submit' action sometimes happens outside of genui, triggered by non-genui ui, that wants user input, even if it is not submitted yet.
2. I am not providing agent to GenUiManager, but handling data flow manually, sometimes directly, sometimes via backend.

Any chance you can accept this PR: https://github.com/flutter/genui/pull/482
Maybe it makes sense to create class GenUiWidgetBuilderInput, to wrap all these fields, to make such changes easier in future. Let me know if you like this alternative more.

jacobsimionato:
Hey this is interesting! I think my position is that we should try to build the API so that it is very rare that someone would need to know SurfaceId within a specific widget. Ideally, we will leak less complexity into CatalogItem rather than more, because CatalogItem is already an overly complex API as you say. But I can see how it could be useful for advanced use cases like yours.

I'm very much in favor of adding GenUiWidgetBuilderInput as you say. Perhaps it could be called something like CatalogItemContext. That would make CLs like this much much easier.

jacobsimionato:
Fixed by Polina's PR

PR #491, opened 2025-11-03
Create data class CatalogItemContext and provide surfaceId to widgetBuilder.

polina-c:
It is follow up for https://github.com/flutter/genui/issues/481

It is successor of https://github.com/flutter/genui/pull/482

Ignoring gemini recommendations as I want to merge this massive change faster, to avoid merge conflicts.

### DIFF
--- a/packages/flutter_genui/lib/src/model/catalog_item.dart
+++ b/packages/flutter_genui/lib/src/model/catalog_item.dart
@@ -23,7 +23,7 @@ typedef ChildBuilderCallback =
 typedef ExampleBuilderCallback = String Function();
 
 /// A callback that builds a widget for a catalog item.
-typedef CatalogWidgetBuilder = Widget Function(CatalogItemContext context);
+typedef CatalogWidgetBuilder = Widget Function(CatalogItemContext itemContext);
 
 class CatalogItemContext {
   CatalogItemContext({


examples/travel_app/lib/src/catalog/checkbox_filter_chips_input.dart
```diff
@@ -101,62 +101,55 @@ final checkboxFilterChipsInput = CatalogItem(
...
-        required id,
-        required buildChild,
-        required dispatchEvent,
-        required context,
-        required dataContext,
-        required getComponent,
-      }) {
-        final checkboxFilterChipsData = _CheckboxFilterChipsInputData.fromMap(
-          data as Map<String, Object?>,
+  widgetBuilder: (context) {
```
gspencergoog, line 104: Because `context` usually refers to a `BuildContext`, can we change the name of this to be "catalogItemContext" or something else that is specific so people reading this can tell it's not a `BuildContext`?

I think this will also help later where you have to come up with other names for the build context: you can go back to calling it `context`, and the catalog item one can be something else.  Maybe `itemContext`?

polina-c: itemContext sounds good!

polina-c: I renamed it in main signature

will rename in other places in next PR

jacobsimionato: +1 to renaming to itemContext. Happy for it to be in another PR, though if both PRs touch the same 50 files, then maybe it make sense to combine them?

PR #493, opened 2025-11-03
Cleanup.

polina-c:


packages/flutter_genui/lib/src/catalog/core_widgets/list.dart
```diff
@@ -77,10 +77,10 @@ final list = CatalogItem(
               scrollDirection: direction,
               itemCount: values.length,
               itemBuilder: (bcontext, index) {
```
gspencergoog: This can go back to being called "context".  And anywhere else you had to rename it to avoid a conflict.

packages/flutter_genui/lib/src/catalog/core_widgets/multiple_choice.dart
```diff
@@ -73,7 +73,7 @@ final multipleChoice = CatalogItem(
       builder: (bcontext, selections, child) {
```
gspencergoog: Can be `context` now.

gspencergoog (approved):

PR #546, opened 2025-11-18
Make it easier to run the travel app and simple chat app on iOS, MacOS and Android without full Firebase setup

jacobsimionato:
This fixes multiple problems:
- Xcode stuff was not checked in which meant hitting a number of errors when running for iOS and MacOS for the first time, e.g. incompatible minimum OS version etc. To fix this, you need to hack the Xcode workspace locally. Now that workspace is fixed and checked in.
- The GoogleServices files are used by Firebase native plugin initialization, and they are not checked in (correctly), so if you try to build the apps without setting up Firebase, it fails looking for the files. Even if you try to run the simple chat app with the Google Generative AI package, you still hit this because native Firebase plugin initialization expects these files. The fix here is to update the stubbing script so that it also creates stubbed GoogleServices files which satisfy the initialization requirements.

Much of this stemmed from the fact that we .gitignored more files than are part of the standard Flutter .gitignore. This change reverts our .gitignores to reflect the standard Flutter tool .gitignore used for new projects, which I assume follows best practices. This means checking in podfile.lock etc.


sethladd:
Thanks! Do we need to update any README to reference/instruct how to use stub_firebase script?

gspencergoog:
Closing this in favor of https://github.com/flutter/genui/pull/544 https://github.com/flutter/genui/pull/547 and https://github.com/flutter/genui/pull/549

PR #549, opened 2025-11-18
Update firebase instructions

gspencergoog:
# Description

This updates the firebase options scripting and the instructions for using them.

I've eliminated the stub files, since you can't actually run anything with those, and updated the refresh_firebase.sh script to take an argument of the firebase project ID, and updated instructions on how to generate the firebase_options.dart file to the README.  I did leave the `firebase_options_stub.dart` files and the `stub_firebase_options.sh` script, since if you don't want to run the examples, but you also don't want to see analyzer errors, you need something.

### DIFF
--- a/examples/simple_chat/lib/main.dart
+++ b/examples/simple_chat/lib/main.dart
@@ -12,16 +12,17 @@ import 'package:genui_google_generative_ai/genui_google_generative_ai.dart';
 import 'package:logging/logging.dart';
 
 import 'configuration.dart';
+
 // If you want to convert to using Firebase AI, run:
 //
 //   sh tool/refresh_firebase.sh <project_id>
 //
 // to refresh the Firebase configuration for a specific Firebase project.
 // and uncomment the Firebase initialization code and import below that is
-// marked with UNCOMMENT_FOR_FIREBASE.
+// marked with UNCOMMENT_FOR_FIREBASE, and set the value of `aiBackend` to
+// `AiBackend.firebase` in `lib/configuration.dart`.
 
-// UNCOMMENT_FOR_FIREBASE
-import 'firebase_options.dart';
+// import 'firebase_options.dart'; // UNCOMMENT_FOR_FIREBASE
 
 // Conditionally import non-web version so we can read from shell env vars in
 // non-web version.
@@ -31,11 +32,11 @@ import 'message.dart';
 void main() async {
   WidgetsFlutterBinding.ensureInitialized();
 
-  // Only initialize Firebase if using firebase backend.
+  // Only initialize Firebase if we are using the Firebase backend.
   if (aiBackend == AiBackend.firebase) {
     await Firebase.initializeApp(
-      // UNCOMMENT_FOR_FIREBASE
-      options: DefaultFirebaseOptions.currentPlatform,
+      // UNCOMMENT_FOR_FIREBASE (See top of file for details)
+      // options: DefaultFirebaseOptions.currentPlatform,
     );
   }
 
--- a/.github/workflows/flutter_packages.yaml
+++ b/.github/workflows/flutter_packages.yaml
@@ -121,8 +121,6 @@ jobs:
       - name: Update submodules
         working-directory: ${{ matrix.package.path }}
         run: git submodule update --init --recursive
-      - name: Stub firebase_options.dart
-        run: sh tool/stub_firebase_options.sh
       - name: Install dependencies
         working-directory: ${{ matrix.package.path }}
         run: dart pub get
--- a/examples/simple_chat/README.md
+++ b/examples/simple_chat/README.md
@@ -69,7 +69,7 @@ This example supports two AI backends: **Google Generative AI** (default) and **
    - Set up a Firebase project
    - Generate a `firebase_options.dart` file using the FlutterFire CLI. You can run `sh tool/refresh_firebase.sh <project_id>` from the repo root to help you set this up.
 
-2. **Switch the Backend**: In `lib/main.dart`, change the
+2. **Switch the Backend**: In `lib/configuration.dart`, change the
    `aiBackend` constant to use Firebase:
 
    ```dart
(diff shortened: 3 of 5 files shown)


gspencergoog:
@jacobsimionato I removed the stub files you added because I didn't see the point of them (since you can't run firebase without the real thing), but maybe I just don't understand how you meant it to work.

Another alternative would be to just check in the stub files already in place (with the real names, e.g. `firebase_options.dart`), and then continue to `.gitignore` them.  That way, changes to them would be ignored, but they would exist when someone clones the repo, and you wouldn't have analyzer errors from the first sync, but you also wouldn't be able to run firebase AI.

I kind of don't mind the analyzer errors because it makes people investigate and either generate the firebase options, or copy over the stubs to get rid of them.  But it is kind of ugly.

jacobsimionato (commented):
For the simple chat app, I think we still want a way of stubbing Firebase so that developers can use the `[genui_google_generative_ai](https://github.com/flutter/genui/tree/main/packages/genui_google_generative_ai)` package with just an API key from AI studio, without needing to go through all the steps of creating a Firebase project, installing local Firebase tooling etc. The stubbed `google-services` and `firebase_options.dart` allow that (though I am also repulsed by the need to run a magical script!).

Another option could be to update simple_chat so that it *only* uses `genui_google_generative_ai`, but then it's harder for people to productionize with Firebase AI Logic. Maybe it's not so bad though. They need to follow all the instructions to get Firebase configured anyway, and then it's only a handful of lines to set up the Firebase content generator. Maybe we could just comment out the lines required to switch to Firebase throughout the codebase. So the developer journey is like:
1. Clone repo
2. Try out simple chat using just an API key
3. Decide to customize productionize app
4. Uncomment some lines to switch to Firebase AI Logic integration
5. Do manual steps to set up Firebase project, run firebase configuration tool etc.
6. Start running app with Firebase AI backend

gspencergoog:
Maybe we can use conditional compilation/import?  Like if you do `flutter run -d macos -D "GEMINI_API_KEY=<api key>"` then it uses `genui_google_generative_ai`, but if you don't, then it uses `firebase_ai`.  The down side of that is that it doesn't use the api key by default.  Maybe we can do it so that if you want to use firebase then you have to define `-D USE_FIREBASE_AI=true` or something?  That way we won't have to have commented out code, and it's easy to switch (and test).

I'll try it out and see if it works.

examples/simple_chat/lib/main.dart
```diff
@@ -11,31 +11,30 @@ import 'package:genui_firebase_ai/genui_firebase_ai.dart';
...
+// If you want to convert to using Firebase AI, run:
+//
+//   sh tool/refresh_firebase.sh <project_id>
+//
+// to refresh the Firebase configuration for a specific Firebase project.
+// and uncomment the Firebase initialization code and import below that is
+// marked with UNCOMMENT_FOR_FIREBASE.
+
+// UNCOMMENT_FOR_FIREBASE
+import 'firebase_options.dart';
```
jacobsimionato: Comment this :-D ?

examples/simple_chat/lib/configuration.dart
```diff
@@ -0,0 +1,16 @@
+// Copyright 2025 The Flutter Authors.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// Enum for selecting which AI backend to use.
+enum AiBackend {
```
jacobsimionato, line 6: It seems like we're now going to have two parallel approaches to switching between backends - this control flow approach with an enum, and directly commenting and uncommenting code.

How about we remove this and *just* use the uncommenting approach to avoid confusion?

examples/simple_chat/lib/main.dart
```diff
@@ -11,31 +11,30 @@ import 'package:genui_firebase_ai/genui_firebase_ai.dart';
...
-
-/// Configuration for which AI backend to use.
-/// Change this value to switch between backends.
-const AiBackend aiBackend = AiBackend.googleGenerativeAi;
-
 void main() async {
   WidgetsFlutterBinding.ensureInitialized();
 
-  // Only initialize Firebase if using firebase backend
+  // Only initialize Firebase if using firebase backend.
```
jacobsimionato: Comment the below?

examples/travel_app/lib/src/config/configuration.dart
```diff
@@ -0,0 +1,16 @@
+// Copyright 2025 The Flutter Authors.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// Enum for selecting which AI backend to use.
+enum AiBackend {
```
jacobsimionato, line 6: This seems inconsistent with the readme which says you *must* use Firebase AI logic. I'm fine either way, but let's choose one or the other and document it, and also perhaps settle on the commenting/uncommenting approach.

tool/stub_firebase_options.sh
jacobsimionato, line 1: Love to see this disappear!

examples/travel_app/lib/src/config/configuration.dart
gspencergoog: Oh, I need to update the README.  Will do.

tool/stub_firebase_options.sh
gspencergoog: Gone!

examples/simple_chat/lib/main.dart
gspencergoog: Oh, I did, but I guess I can remove the blank line above it so the comment is associated with that line.

examples/simple_chat/lib/configuration.dart
gspencergoog: Yes, good point.  I was leaving it there so that you could still switch between the two if after configuring Firebase AI if you wanted to, but I could remove that mechanism if you think that's better.

gspencergoog: Actually, the other reason to leave it so that you only need to uncomment two lines instead of several lines in multiple files.  I think I'd like to just keep both methods.

jacobsimionato (approved):

PR #571, opened 2025-11-24
Add documentation for the a2ui protocol message lifecycle

jacobsimionato:
As requested by some customers, this adds some docs on the A2UI protocol. This will be unnecessary once A2UI has better public documentation, but this is to help in the mean time.


Piinks:
Greetings from stale PR triage! 👋 
Is this change still on your radar?

Issue #570, opened 2025-11-22
Dartantic ContentGenerator for the GenUI package


sethladd:
### Discussed in https://github.com/flutter/genui/discussions/524

<div type='discussions-op-text'>

<sup>Originally posted by **csells** November 14, 2025</sup>
If someone tells me what this means, I'll happily build one. It sounds cool. : )</div>

PR #583, opened 2025-12-09
Add Dartantic Content Generator

csells:
Adds `genui_dartantic` package - a new `ContentGenerator` implementation that enables GenUI to work with multiple AI providers (OpenAI, Anthropic, Google, Mistral, Cohere, Ollama) via the dartantic_ai package.

## Issues fixed
- https://github.com/flutter/genui/issues/570

## Elements of the package

- DartanticContentGenerator
- DartanticContentConverter
- DartanticSchemaAdapter
- Simplified Travel Planner Example App
  - including the complete catalog -- should that come from somewhere else?

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/packages/genui_dartantic/lib/src/dartantic_content_generator.dart
+++ b/packages/genui_dartantic/lib/src/dartantic_content_generator.dart
@@ -64,6 +64,20 @@ class DartanticContentGenerator implements ContentGenerator {
       chatModelName: modelName,
       tools: dartanticTools,
     );
+
+    // Create additional system instructions to augment what the client sends
+    _extraInstructions =
+        '''
+<tools>
+${dartanticTools.map((tool) => tool.toJson()).join('\n\n')}
+</tools>
+
+<output_schema>
+${_outputSchema.toJson()}
+</output_schema>
+''';
+
+    genUiLogger.info('Extra system instructions: $_extraInstructions');
   }
 
   /// The catalog of UI components available to the AI.
@@ -84,6 +98,7 @@ class DartanticContentGenerator implements ContentGenerator {
   final _textResponseController = StreamController<String>.broadcast();
   final _errorController = StreamController<ContentGeneratorError>.broadcast();
   final _isProcessing = ValueNotifier<bool>(false);
+  late final String _extraInstructions;
 
   /// Structured output schema: a simple object with a required string response.
   static final JsonSchema _outputSchema = JsonSchema.create({
@@ -128,7 +143,7 @@ class DartanticContentGenerator implements ContentGenerator {
       // Convert GenUI history to dartantic ChatMessage list
       final List<dartantic.ChatMessage> dartanticHistory = _converter.toHistory(
         history,
-        systemInstruction: systemInstruction,
+        systemInstruction: '$systemInstruction\n\n$_extraInstructions',
       );
 
       // Convert the current GenUI message into prompt text plus parts so we
--- a/packages/genui_dartantic/example/lib/src/game_page.dart
+++ b/packages/genui_dartantic/example/lib/src/game_page.dart
@@ -36,15 +36,17 @@ class _GamePageState extends State<GamePage> {
 
   void _updateThinkingVerb() {
     if (_conversation.isProcessing.value) {
-      setState(() {
-        _thinkingVerb = thinkingVerbs[Random().nextInt(thinkingVerbs.length)];
-      });
+      setState(
+        () => _thinkingVerb =
+            thinkingVerbs[Random().nextInt(thinkingVerbs.length)],
+      );
     }
   }
 
   @override
   void initState() {
     super.initState();
+
     _genUiManager = A2uiMessageProcessor(catalogs: [ticTacToeCatalog]);
     _conversation = GenUiConversation(
       contentGenerator: widget.generator,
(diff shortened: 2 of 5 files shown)


packages/genui_dartantic/example/lib/src/catalog/itinerary.dart
```diff
@@ -0,0 +1,643 @@
...
+                          onPressed: () {
+                            final JsonMap? actionData =
+                                data.choiceRequiredAction;
+                            if (actionData == null) {
+                              return;
+                            }
+                            final actionName = actionData['name'] as String;
+                            final List<Object?> contextDefinition =
+                                (actionData['context'] as List<Object?>?) ??
+                                <Object>[];
```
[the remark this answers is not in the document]
csells: @gemini-code-assist  can you make a suggestion?

packages/genui_dartantic/example/lib/src/catalog/listings_booker.dart
```diff
@@ -0,0 +1,536 @@
...
+                    shape: BoxShape.circle,
+                    color: Theme.of(context).primaryColor,
+                  ),
+                ),
+              )
+            : null,
+      ),
+    );
+  }
+}
```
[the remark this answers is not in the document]
csells: @gemini-code-assist can you make a suggestion?

parlough:
Exciting @csells! :D 

I'm not familiar with the setup here, but should this be separate from this repo so we can point to it as an exemplary example of a community-maintained provider/generator?

csells:
Seth and Greg asked for a PR but making this an external repo works for me, too.

sethladd:
thanks @csells ! I think it's OK to keep in this repo now. We might split it out later, once APIs stabilize.

parlough:
Keeping it here sounds good to me! Thanks for clarifying :)

packages/genui_dartantic/lib/src/dartantic_content_generator.dart
```diff
@@ -0,0 +1,218 @@
...
+
+// ignore_for_file: specify_nonobvious_local_variable_types
+
+import 'dart:async';
+
+import 'package:dartantic_ai/dartantic_ai.dart' as dartantic;
+import 'package:dartantic_interface/dartantic_interface.dart' as di;
+import 'package:flutter/foundation.dart';
+import 'package:genui/genui.dart';
+import 'package:json_schema/json_schema.dart';
```
gspencergoog, line 12: Ideally, this would use [json_schema_builder](https://github.com/flutter/genui/tree/main/packages/json_schema_builder) instead of json_schema, since all of the other content generators use it, but I guess this is required by dartantic?

packages/genui_dartantic/lib/src/dartantic_content_generator.dart
```diff
@@ -0,0 +1,218 @@
...
+  late final dartantic.Agent _agent;
+  final DartanticContentConverter _converter = DartanticContentConverter();
+
+  final _a2uiMessageController = StreamController<A2uiMessage>.broadcast();
+  final _textResponseController = StreamController<String>.broadcast();
+  final _errorController = StreamController<ContentGeneratorError>.broadcast();
+  final _isProcessing = ValueNotifier<bool>(false);
+
+  /// Structured output schema: a simple object with a required string response.
+  static final JsonSchema _outputSchema = JsonSchema.create({
```
gspencergoog, line 104: This is going to create text responses embedded in JSON, not JSON output, correct?  I'm not sure what the point of this is: why use structured output to get text responses unless you're using forced function calling?

packages/genui_dartantic/lib/src/dartantic_content_generator.dart
```diff
@@ -0,0 +1,218 @@
...
+
+      // We should never have tool calls or results in request message.
+      assert(promptAndParts.parts.every((part) => part is! di.ToolPart));
+
+      genUiLogger.info(
+        'Sending request to Dartantic: "${promptAndParts.prompt}"',
+      );
+      genUiLogger.fine('History contains ${dartanticHistory.length} messages');
+
+      // Use Agent.sendFor with structured output so the model returns a single
```
gspencergoog, line 162: What does this do for models that don't support structured output? Does it just not work there, or does it do some processing to try and capture the JSON from the text output or tool inputs?

packages/genui_dartantic/lib/src/dartantic_schema_adapter.dart
```diff
@@ -0,0 +1,395 @@
...
+        );
+        if (adaptedProperty != null) {
+          properties[entry.key] = adaptedProperty;
+        }
+      }
+    }
+
+    _warnIfPresent(objectSchema.patternProperties, 'patternProperties', path);
+    _warnIfPresent(objectSchema.dependentRequired, 'dependentRequired', path);
+    _warnIfPresent(
```
gspencergoog: Isn't `additionalProperties: false` allowed?  I think on some AI platforms (OpenAI), it's actually required (and required to be false), right?

packages/genui_dartantic/example/lib/main.dart
gspencergoog, line 1: I'm not sure it makes sense to duplicate the entire `travel_app` example here.  It would probably make more sense to have a separate, simpler, example for dartantic. Something that doesn't require lots of assets, with less code to maintain would be ideal.  The idea isn't to have an example of GenUI (that's what the top level examples are for), but rather to have an example of what is different about using Dartantic from the other content generators.

packages/genui_dartantic/lib/src/dartantic_content_generator.dart
csells: as discussed offline, dartantic uses json_schema.

packages/genui_dartantic/lib/src/dartantic_content_generator.dart
csells: as discussed offline, dartantic handles structured output and tool calls in the same call, so doesn't require the "final result" style work-around.

packages/genui_dartantic/lib/src/dartantic_content_generator.dart
csells: as discussed offline, this gives back structured JSON (Map<String, dynamic>), not a JSON string.

packages/genui_dartantic/lib/src/dartantic_schema_adapter.dart
csells: yes it is and I'll fix that.

csells: removed all of that warn stuff -- none of it is needed. removed about 300 loc.

packages/genui_dartantic/example/lib/main.dart
csells: replaced the travel app with a tic-tac-toe game that allows you to choose the provider to play against. OpenAI doesn't work right now, but I can make it work by moving to dartantic 2.0. that move [requires a2a to upgrade their use of mcp_dart](https://github.com/shamblett/a2a/issues/60).

gspencergoog (approved):

gspencergoog:
You can fix most of the errors there by just running `./tool/run_all_tests_and_fixes.sh` which should fix the copyrights and any formatting issues.

csells:
@gspencergoog this is ready to be merged unless you want to wait for dartantic v2, which should make anthropic work better and openai work at all. gemini flash already works well.

sethladd:
Curious how the tic tac toe game shows off the power of GenUI in an agentic conversational collaboration? I admit I haven't run it yet.

csells:
@sethladd it generates the UI for every turn.

PR #587, opened 2025-12-09
Rename GenUiManager to A2uiMessageProcessor

jacobsimionato:
This is to be consistent with A2UI naming - see https://github.com/google/A2UI/issues/163


jacobsimionato:
I have addressed the feedback. Please take another look.

Issue #559, opened 2025-11-19
GenUI TextField crashes with “InputDecorator unbounded width” when generated using A2UI (Flutter GenUI v0.5.1)


SheikMohideen007:
Use this system prompt (or any prompt that generates TextFields inside a Row/Column):

Your job is to build UI using ONLY the built-in CoreCatalogItems.

Always generate:
1. A top search bar (TextField + Button)
2. A search results list
3. A chat message list
4. A bottom input area (TextField + Button)

IMPORTANT:
- Wrap every TextField in a width-bounded Box or Expanded.
- Do NOT put TextField directly inside a Row or Column without constraints.
- Do not specify surface IDs manually.
- Always send BeginRendering after updates.

=====================================================================================
Actual Results :

Flutter throws the runtime error:
InputDecorator cannot have an unbounded width.
TextField must be inside a width-bounded parent.

The crash happens even when:

1. Instructions explicitly force width constraints

2. TextFields are wrapped in Containers, Boxes, or Expanded

3. No manual surface IDs are used
 
4. Using only default CoreCatalogItems

=====================================================================================
Expected Results :
GenUI should generate TextField layouts that obey Flutter’s width-bounded constraint:

1. TextField inside Box(width: infinity)

2. TextField inside Expanded

3. TextField inside Container (width constraints)

4. The UI should render without crashing.

=====================================================================================
Sample Code:

import 'package:flutter/material.dart';
import 'package:genui/genui.dart';
import 'package:genui_firebase_ai/genui_firebase_ai.dart';

class Home extends StatefulWidget {
  @override
  State<Home> createState() => _HomeState();
}

class _HomeState extends State<Home> {
  late GenUiConversation conversation;
  final _surfaceIds = <String>[];

  @override
  void initState() {
    super.initState();

    final catalog = CoreCatalogItems.asCatalog();

    final generator = FirebaseAiContentGenerator(
      catalog: catalog,
      systemInstruction: "<SYSTEM PROMPT FROM ABOVE>",
    );

    conversation = GenUiConversation(
      genUiManager: GenUiManager(catalog: catalog),
      contentGenerator: generator,
      onSurfaceAdded: (s) => setState(() => _surfaceIds.add(s.surfaceId)),
      onSurfaceDeleted: (s) => setState(() => _surfaceIds.remove(s.surfaceId)),
    );

    generator.a2uiMessageStream.listen((msg) {
      conversation.genUiManager.handleMessage(msg);
    });
  }

  @override
[...]

PR #588, opened 2025-12-09
fix(genui): missing `weight` property to `Component` constructor; Prevent unbounded text field width

andrewkolos:
## Description

This is a two-for-one change. Feel free to ask me to refactor this into two PRs.

Fixes https://github.com/flutter/genui/issues/559. This isn't an air-tight fix, but I hope it is sufficient for now.

## Changes 
* `SurfaceUpdateTool` wasn't parsing the value of the `weight` property and passing it to the `Component` constructor for any component it was constructing[^1]. Definition of `weight` property: https://github.com/flutter/genui/blob/061ba8943160894a16ab0ffeaae99b7859e0ac49/packages/genui/lib/src/model/a2ui_schemas.dart#L227-L230.
* The `widgetBuilder` of the text field catalog component was modified. It now uses `LayoutBuilder` to determine if whether the width is unbounded, in which case a warning will be logged and the text field will be rendered in a `SizedBox` with a width of 300 pixels. I chose this width arbitrarily.
* **Hack:** In the schema, the `description` of `TextField` and the `description` of `weight` now include a specific call out about `TextField` and how it needs its width bound. This creates an undesirable coupling between `TextField` and `Component`, and it feels wrong for `Component` to be aware of a downstream detail like `TextField`. However, this was the only way I found to get Gemini Flash 2.5 to respect this rule with any level of consistency (Pro 2.5 seems to be more rigorous and anticipate the need to use `weight` without explicit instruction). Making these explicit callouts may also impose more token burden on Flash (and similarly lightweight) models, possibility reducing the quality of the library's output elsewhere, but we currently do not have a way to measure this.

## Repro
You can checkout this branch and create a basic test app with the following (LLM-generated) code:

<details>
<summary> Repro app code </summary>

```dart
import 'package:flutter/material.dart';
import 'package:genui/genui.dart';
import 'package:genui_google_generative_ai/genui_google_generative_ai.dart';

void main() {
  configureGenUiLogging();
  runApp(const ReproApp());
}

class ReproApp extends StatelessWidget {
  const ReproApp({super.key});

  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'GenUI Repro',
      theme: ThemeData(primarySwatch: Colors.blue),
      home: const Home(),
    );
  }
}

class Home extends StatefulWidget {
  const Home({super.key});

  @override
  State<Home> createState() => _HomeState();
}

class _HomeState extends State<Home> {
  GenUiConversation? conversation;
  final _surfaceIds = <String>[];

  @override
  void initState() {
    super.initState();
    _startConversation();
  }

  void _startConversation() {
    final catalog = CoreCatalogItems.asCatalog();
    final generator = GoogleGenerativeAiContentGenerator(
      catalog: catalog,
      modelName: 'models/gemini-2.5-flash',
      systemInstruction: '''
Your job is to build UI using ONLY the built-in CoreCatalogItems.

Always generate:
1. A top search bar (TextField + Button)
2. A search results list
3. A chat message list
4. A bottom input area (TextField + Button)
''',
    );

    setState(() {
      conversation = GenUiConversation(
        genUiManager: GenUiManager(catalogs: [catalog]),
        contentGenerator: generator,
        onSurfaceAdded: (s) => setState(() => _surfaceIds.add(s.surfaceId)),
        onSurfaceDeleted: (s) =>
            setState(() => _surfaceIds.remove(s.surfaceId)),
        onError: (e) => print('Error: $e'),
      );
    });

    // Auto-trigger the generation
    conversation?.sendRequest(
      UserMessage([TextPart('Generate the UI please')]),
    );
  }

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      appBar: AppBar(
        title: const Text('GenUI Real Repro'),
        actions: [
          IconButton(
            icon: const Icon(Icons.refresh),
            onPressed: () {
              // Reset
              setState(() {
                _surfaceIds.clear();
                conversation?.dispose();
                conversation = null;
              });
              // Give it a moment to dispose before restarting
              Future.delayed(
                const Duration(milliseconds: 100),
                _startConversation,
              );
            },
          ),
        ],
      ),
      body: Column(
        children: [
          Expanded(
            child: ListView.builder(
              itemCount: _surfaceIds.length,
              itemBuilder: (context, index) {
                return GenUiSurface(
                  host: conversation!.host,
                  surfaceId: _surfaceIds[index],
                );
              },
            ),
          ),
          if (conversation != null)
            ValueListenableBuilder(
              valueListenable: conversation!.isProcessing,
              builder: (context, isProcessing, _) {
                if (!isProcessing) return const SizedBox.shrink();
                return const LinearProgressIndicator();
              },
            ),
        ],
      ),
      floatingActionButton: FloatingActionButton(
        onPressed: () {
          conversation?.sendRequest(
            UserMessage([TextPart('Generate the UI please')]),
          );
        },
        child: const Icon(Icons.play_arrow),
      ),
    );
  }
}
```
</details>

## Pre-launch Checklist

- [x] I signed the [CLA].

[changelog]: ../CHANGELOG.md

[^1]: https://github.com/flutter/genui/blob/061ba8943160894a16ab0ffeaae99b7859e0ac49/packages/genui/lib/src/core/ui_tools.dart#L35-L37

### DIFF
--- a/packages/genui/test/catalog/core_widgets/text_field_test.dart
+++ b/packages/genui/test/catalog/core_widgets/text_field_test.dart
@@ -11,9 +11,7 @@ void main() {
   testWidgets('TextField in Row with no weight uses a static default width', (
     WidgetTester tester,
   ) async {
-    final manager = GenUiManager(
-      catalogs: [CoreCatalogItems.asCatalog()],
-    );
+    final manager = GenUiManager(catalogs: [CoreCatalogItems.asCatalog()]);
     const surfaceId = 'testSurface';
     final components = [
       const Component(
@@ -66,9 +64,7 @@ void main() {
   testWidgets('TextField in Row (with weight) expands', (
     WidgetTester tester,
   ) async {
-    final manager = GenUiManager(
-      catalogs: [CoreCatalogItems.asCatalog()],
-    );
+    final manager = GenUiManager(catalogs: [CoreCatalogItems.asCatalog()]);
     const surfaceId = 'testSurface';
     final components = [
       const Component(


packages/genui/lib/src/catalog/core_widgets/text_field.dart
```diff
@@ -10,9 +10,18 @@ import '../../model/a2ui_schemas.dart';
...
+import '../../primitives/logging.dart';
 import '../../primitives/simple_items.dart';
 
+/// The default width used when the TextField is unbounded.
+@visibleForTesting
+const double kDefaultTextFieldWidth = 300.0;
+
 final _schema = S.object(
+  description:
+      'A text input field. IMPORTANT: If this is placed inside a Row, '
```
jacobsimionato, line 22: Hmmm I think we can't rely on this, because weight should always be optional. If I put a textField inside a row, shouldn't the row have a maximum width, and the textField be constrained to that? I know that is clearly not what's happening here, but that's what I'd hope for.

Can we do something like default the weight to 1.0 essentially, even if it isn't specified? 

The overall goal here is to provide enough default behavior that the LLM can compose any element inside any other without complying with additional rules. The LLM here may be running on a server and not know the difference between flutter or lit etc, so we can't impose platform-specific constraints!

jacobsimionato (changes requested):

packages/genui/lib/src/catalog/core_widgets/text_field.dart
andrewkolos: Thanks for the healthy scrutinization here! I've incorporated your suggestion (and some of my thoughts about it) in a new PR that will supersede this one: https://github.com/flutter/genui/pull/603. I chose to make a new PR as this idea is fairly different than the one I proposed in this one.

andrewkolos:
Superseded by https://github.com/flutter/genui/pull/603.

PR #591, opened 2025-12-10
chore(lints): `remove specify_nonobvious_local_variable_types` rule

andrewkolos:
## Description

`specify_nonobvious_local_variable_types` asks us to write/read explicit type declarations for lines like

``` dart
final localizations = MaterialLocalizations.of(context);
```

and

```dart
final date = DateTime.parse(value);
```

In these scenarios, I find the variable data type to be quite obvious. Since this rule isn't programmable to not trip in these scenarios, I propose we remove it and trust code authors to use or omit types at their discretion.

## Pre-launch Checklist

- [x] I signed the [CLA].

[changelog]: ../CHANGELOG.md

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -40,5 +40,4 @@ linter:
     - unreachable_from_main
 
     - omit_obvious_local_variable_types
-    - specify_nonobvious_local_variable_types
     - specify_nonobvious_property_types


gspencergoog:
I added that to be consistent with the flutter/flutter repo, since we're part of the flutter repo and project.

If others disagree with it, we can remove it.  I agree that in the case of factory constructors the type can be a little obvious, but in other cases, it's quite useful to see the type, especially for developers who don't know the code, and to drive it home for LLMs inspecting the code.

polina-c:
> I added that to be consistent with the flutter/flutter repo, since we're part of the flutter repo and project.

@gspencergoog , what is the reason we are itemizing lints, instead of using just dart_flutter_team_lints? If there is one, should we document it here in a comment?

Yes, this lint is sometimes helpful and sometimes noisy. If `dart fix` auto-fixes the issues, I would keep it. If no, I vote to remove.

andrewkolos:
> I added that to be consistent with the flutter/flutter repo, since we're part of the flutter repo and project.

Ah okay thanks for the context. I didn't recall that flutter/flutter has this enabled. I'm okay with keeping it as the cost is low enough.

> If dart fix auto-fixes the issues, I would keep it. If no, I vote to remove.

`dart fix` does indeed work for this lint.

andrewkolos:
Closing. The net benefit of removing this lint is ambiguous.

gspencergoog:
> > I added that to be consistent with the flutter/flutter repo, since we're part of the flutter repo and project.
> 
> @gspencergoog , what is the reason we are itemizing lints, instead of using just dart_flutter_team_lints? If there is one, should we document it here in a comment?

Because `dart_flutter_team_lints` doesn't match the flutter/flutter lints.

polina-c:
> > > I added that to be consistent with the flutter/flutter repo, since we're part of the flutter repo and project.
> > 
> > 
> > @gspencergoog , what is the reason we are itemizing lints, instead of using just dart_flutter_team_lints? If there is one, should we document it here in a comment?
> 
> Because `dart_flutter_team_lints` doesn't match the flutter/flutter lints.

Why we choose flutter/flutter lints over dart_flutter_team_lints?

`dart_flutter_team_lints` seem to be faster to pickup good things, no?

And why we are itemizing lints, instead of just referencing the package?

PR #603, opened 2025-12-15
fix(genui): add missing `weight` property to `Component` constructor; default `TextField` `width` to 1 when nested in a `Row`

### LINKED ISSUE #559: GenUI TextField crashes with “InputDecorator unbounded width” when generated using A2UI (Flutter GenUI v0.5.1)

andrewkolos:
## Description

Supersedes https://github.com/flutter/genui/pull/588. Fixes #559.

## Changes 
* **fix 1: `SurfaceUpdateTool`** wasn't parsing the value of the `weight` property and passing it to the `Component` constructor for any component it was constructing[^1]. Definition of `weight` property: https://github.com/flutter/genui/blob/061ba8943160894a16ab0ffeaae99b7859e0ac49/packages/genui/lib/src/model/a2ui_schemas.dart#L227-L230. This change fixes this and makes sure the weight value gets plumbed through properly.
  * I will also use this opportunity to say that optional fields should be largely avoided in internal APIs. Optional parameters are very prone to situations like these where code authors forget to pass values when they meant to (or simply never discovered the parameter's existence to begin with). Optional parameters are best reserved for highly flexible user-facing libraries that need to be both very simple to start with but also highly configurable (e.g. components in UI frameworks, such as `Text` in Flutter). Since we are dealing with an internal API here that is not suffering from any significant amount of API bloat, this PR opts to make `weight` parameter required. The only material drawback of making this parameter required is some extra verbosity in test code, which I personally assign very little importance to.
* **(necessary) refactoring:** `buildWeightedChild` no longer has a `required Component? component` parameter and instead has a `required int? weight` parameter. `component` was only being read for the weight value anyway, and this refactoring also lets us "override" the weight of a `component`. Continue reading to see why we need this.
* **(hacky) fix 2:** the builder of the row catalog item now checks the type of each child component. If it sees a `TextField` without a weight, it will assign it a weight of `1`. This will prevent the layout-induced crashes we see in #559. However, I call this hacky because 1) Row has to be explicitly aware of another catalog component and treat it specially and 2) this is essentially a concession that many A2UI layouts are doomed to be laid out in noticeably different ways across different renderers (even if you tweaked the common variables like colors and padding to be the same across each of the renderers). I am okay with making this concession for now, especially that different UI frameworks fundamentally compute constraints in differing ways, making this a high-cost (near intractable?) problem to solve. It also isn't clear to me how many software products will really suffer from the problem to a significant degree (e.g. "we need a piece of generated UI to _look_ identical on both our HTML and native Android apps").

## Screenshots

image

Created using the repro instructions provided by the parent issue. In this scenario, the agent did not provide any weights for any child elements of rows, so this PR's change prevents a crash by assigning a `weight` of `1` to the text fields.

Another example (using simple_chat):

image

## Pre-launch Checklist

- [x] I signed the [CLA].

[changelog]: ../CHANGELOG.md

[^1]: https://github.com/flutter/genui/blob/061ba8943160894a16ab0ffeaae99b7859e0ac49/packages/genui/lib/src/core/ui_tools.dart#L35-L37

### DIFF
--- a/packages/genui/lib/src/catalog/core_widgets/row.dart
+++ b/packages/genui/lib/src/catalog/core_widgets/row.dart
@@ -8,6 +8,7 @@ import 'package:json_schema_builder/json_schema_builder.dart';
 import '../../model/a2ui_schemas.dart';
 import '../../model/catalog_item.dart';
 import '../../model/data_model.dart';
+import '../../model/ui_models.dart';
 import '../../primitives/simple_items.dart';
 import 'widget_helpers.dart';
 
@@ -134,6 +135,10 @@ final row = CatalogItem(
         );
       },
       templateListWidgetBuilder: (context, list, componentId, dataBinding) {
+        final Component? component = itemContext.getComponent(componentId);
+        final int? weight =
+            component?.weight ?? (component?.type == 'TextField' ? 1 : null);
+
         return Row(
           mainAxisAlignment: _parseMainAxisAlignment(rowData.distribution),
           crossAxisAlignment: _parseCrossAxisAlignment(rowData.alignment),
@@ -146,11 +151,7 @@ final row = CatalogItem(
                   DataPath('$dataBinding/$i'),
                 ),
                 buildChild: itemContext.buildChild,
-                weight:
-                    itemContext.getComponent(componentId)?.weight ??
-                    (itemContext.getComponent(componentId)?.type == 'TextField'
-                        ? 1
-                        : null),
+                weight: weight,
               ),
             ],
           ],


packages/genui/lib/src/catalog/core_widgets/row.dart
```diff
@@ -142,7 +146,11 @@ final row = CatalogItem(
                   DataPath('$dataBinding/$i'),
                 ),
                 buildChild: itemContext.buildChild,
-                component: itemContext.getComponent(componentId),
+                weight:
+                    itemContext.getComponent(componentId)?.weight ??
+                    (itemContext.getComponent(componentId)?.type == 'TextField'
+                        ? 1
+                        : null),
```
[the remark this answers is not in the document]
andrewkolos: The suggested code is incorrect, but the sentiment here is good. I incorporated it into [avoid re-fetching weight/type when looping over each child](https://github.com/flutter/genui/pull/603/commits/b2ebc54978fed9f5886e0e0f18cdcabdabbce803)

jacobsimionato:
Hey this sounds very good! Can you include a screenshot with some example rendered layouts?

andrewkolos:
> Hey this sounds very good! Can you include a screenshot with some example rendered layouts?

Added some screenshots.

packages/genui/lib/src/catalog/core_widgets/row.dart
```diff
@@ -123,13 +124,21 @@ final row = CatalogItem(
...
+                          : null),
                 ),
               )
               .toList(),
         );
       },
       templateListWidgetBuilder: (context, list, componentId, dataBinding) {
+        final Component? component = itemContext.getComponent(componentId);
+        final int? weight =
+            component?.weight ?? (component?.type == 'TextField' ? 1 : null);
```
jacobsimionato, line 140: This LGTM as a practical fix for this now! I think we need a more scalable solution long term, to avoid other leaf components having this issue. We should figure out what circumstances this error happens in - is it for any leaf component that has an unbounded max width or height? Maybe we need them to be identifiable somehow. Either way, this is a great first step I think!

jacobsimionato (approved):
Nice!

packages/genui/lib/src/catalog/core_widgets/row.dart
andrewkolos: Yeah, constraints go down; sizes go up and all that[^1]. Basically any widget that relies on a constraint from their parent to figure out how large they should be. For example, I suspect sliders within rows will fall victim to this exact crash.

[^1]: https://docs.flutter.dev/ui/layout/constraints

andrewkolos: And yeah, the cleaner solution would be some sort of declaration within the catalog that describe what constraint(s) a widget needs to be able to render. We'd also have to do some plumbing and/or refactoring to make this information available to layout widgets like `Row` and `Column`.

PR #619, opened 2025-12-18
initial port of dartantic_interface types

csells:
# Description

fixes https://github.com/flutter/genui/issues/626

Adds the following types ported from dartantic_interface and migrated to json_schema_builder:
- Part (base class)
- TextPart
- DataPart (for images/binary data)
- LinkPart (for URLs)
- ToolPart (for tool calls and results)
- ToolPartKind (enum: call, result)
- ChatMessage
- ChatMessageRole (enum: system, user, model)
- ToolDefinition<TInput> (name, description, inputSchema but no call implementation)

## Pre-launch Checklist

- [x] I signed the [CLA].
- [n/a] I have added sample code updates to the [changelog].


csells:
fyi: @gspencergoog @polina-c @sethladd

packages/genai_primitives/lib/src/message_parts.dart
```diff
@@ -0,0 +1,333 @@
...
+      ) =>
+        {
+          'id': id,
+          'name': name,
+          if (arguments != null) 'arguments': arguments,
+          if (result != null) 'result': result,
+        },
+      _ => throw UnimplementedError('Unknown part type: $runtimeType'),
+    },
+  };
```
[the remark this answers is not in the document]
csells: /gemini suggest that fix

packages/genai_primitives/lib/src/message_parts.dart
```diff
@@ -0,0 +1,333 @@
...
+  /// Gets the extension for a MIME type.
+  static String? extensionFromMimeType(String mimeType) {
+    final String ext = defaultExtensionMap.entries
+        .firstWhere(
+          (e) => e.value == mimeType,
+          orElse: () => const MapEntry('', ''),
+        )
+        .key;
+    return ext.isNotEmpty ? ext : null;
+  }
```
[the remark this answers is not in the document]
csells: it's worse to duplicate a half-assed version of the list of extensions that isn't properly maintained

packages/genai_primitives/lib/src/chat_message.dart
```diff
@@ -0,0 +1,133 @@
...
+    Object.hashAll(metadata.entries),
+  );
+
+  @override
+  String toString() =>
+      'Message(role: $role, parts: $parts, metadata: $metadata)';
+}
+
+/// The role of a message author.
+enum ChatMessageRole {
```
polina-c, line 129: Can you add more details to doc strings for enum and its items?

I can represent a confused user here:

- doc for enum is saying that each item is a role of a message author, while doc for each item is describing type of message, not author

- can you elaborate what is 'system'?

- documentation is saying 'from', but not 'to'. Is it correct that system always send messages to model? If yes, should we specify it here? And the same about model: is it correct that model's messages are intended to be handled by system and then system will show to the user whatever is intended for user?

- or my assumptions above are completely wrong and the enum relates to difference between 'user prompt' and 'system prompt' somehow?

packages/genai_primitives/lib/src/message_parts.dart
```diff
@@ -0,0 +1,341 @@
...
+    required this.name,
+    required this.result,
+  }) : kind = ToolPartKind.result,
+       arguments = null;
+
+  /// The kind of tool interaction.
+  final ToolPartKind kind;
+
+  /// The unique identifier for this tool interaction.
+  final String id;
```
polina-c: How about renaming it to `interactionId`, to avoid confusion and make it clear how to name this id, when it is copied outside of this class? 
Is it correct that id of tool call is expected to be equal to id of tool call result? If yes, should it be documented here?

packages/genai_primitives/lib/src/message_parts.dart
```diff
@@ -0,0 +1,341 @@
...
+  /// The result of a tool execution (null for calls).
+  final dynamic result;
+
+  /// The arguments as a JSON string.
+  String get argumentsRaw => arguments != null
+      ? (arguments!.isEmpty ? '{}' : jsonEncode(arguments))
+      : '';
+
+  @override
+  bool operator ==(Object other) =>
```
polina-c: Would you mind updating it to use this boilerplate, so that readers do not have to detect if it is identical to what is recommended in style guide:

https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#common-boilerplates-for-operator--and-hashcode

polina-c:
Thank you! It looks good. Left some naming and documentation related comments.

packages/genai_primitives/lib/src/message_parts.dart
```diff
@@ -0,0 +1,341 @@
...
+       arguments = null;
+
+  /// The kind of tool interaction.
+  final ToolPartKind kind;
+
+  /// The unique identifier for this tool interaction.
+  final String id;
+
+  /// The name of the tool.
+  final String name;
```
polina-c: Should it be named 'toolName'? Unlike other properties 'name', this 'name' is not name of the enclosing object (ToolCall or ToolResult).

csells:
Yes you're correct about the tool result id matching the tool can id. I like the idea of updating the name and the docs to reflect that, although I don't know what that name should be. I'll provide an update ASAP.

packages/genai_primitives/lib/src/message_parts.dart
```diff
@@ -0,0 +1,341 @@
...
+
+  /// The arguments for a tool call (null for results).
+  final Map<String, dynamic>? arguments;
+
+  /// The result of a tool execution (null for calls).
+  final dynamic result;
```
polina-c, line 276: Should we use `Object?` instead of dynamic everywhere?

gspencergoog: I second that.  Dynamic turns off type checking.

polina-c:
> Yes you're correct about the tool result id matching the tool can id. I like the idea of updating the name and the docs to reflect that, although I don't know what that name should be. I'll provide an update ASAP.

I suggest `interactionId`.

packages/genai_primitives/lib/src/message_parts.dart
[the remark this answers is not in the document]
gspencergoog: Agreed.  Really the `mime` package should be fixed to export a reverse lookup.

packages/genai_primitives/lib/src/chat_message.dart
polina-c: Maybe in this library instead of ChatMessage having field `ChatMessageRole role`, it make sense for it to have the field 'String type', so that every dependent can define their set of message types.

polina-c:
@csells,  I raised many questions about minor details. 

There are options how to proceed:
1. I can just merge your PR, then figure out the details in the follow up PRs, then publish the package. 
2. We can figure out the details on this PR. 

If you want to actively participate in discussion, let's go with (2).
Otherwise, let's proceed with (1), where you will be optional reviewer for the follow up PRs.

packages/genai_primitives/lib/src/chat_message.dart
polina-c: Another option:

```dart
class ChatMessage<T>{
  ...
  final T type;
  ...
}
```

Then users will be able to pass their enums for the message type. Maybe some of them will have more than one model and more than one system.

polina-c: Or, we can remove type completely, so that users just derive from ChatMessage and add whatever fields they want to add. Or users can compose ChatMessage into another object, that contains envelope information.

This seems to be cleanest option for me.

polina-c:
> > Yes you're correct about the tool result id matching the tool can id. I like the idea of updating the name and the docs to reflect that, although I don't know what that name should be. I'll provide an update ASAP.
> 
> 
> 
> I suggest `interactionId`.

Or 'requestId'?

packages/genai_primitives/lib/src/chat_message.dart
csells: I'll elaborate on what the enums mean, but across all of the LLM APIs (and I've built against most of them by now, certainly all of the major ones), it always boils down to role (user or system or model|assistant) and parts. Anything more than that is over engineering, imo.

packages/genai_primitives/lib/src/message_parts.dart
csells: yes. that's correct. I'll update the name and the docs.

packages/genai_primitives/lib/src/message_parts.dart
csells: will do

packages/genai_primitives/lib/src/message_parts.dart
csells: good idea. will do.

packages/genai_primitives/lib/src/message_parts.dart
csells: what kind of type checking do we get with Object? the standard JSON in/out in Dart is Map<String, dynamic> -- shouldn't we stick with that? I'm open to changing to Object? if I can understand the reason to vary from the Dart SOP.

packages/genai_primitives/lib/src/chat_message.dart
polina-c: I still think having roles itemized makes this library too opinionated.

I would remove role from the message completely, allowing clients to define the roles in their systems.

If you disagree let's merge this as is and have this discussion outside of this PR.

polina-c:
Can you remove changes in packages/genui_dartantic?

packages/genai_primitives/lib/src/chat_message.dart
csells: let's talk after the merge.

csells:
> Can you remove changes in packages/genui_dartantic?

I needed those changes to get the checks to pass.

packages/genai_primitives/lib/src/message_parts.dart
polina-c: There are debates in Dash team if Dart should change the standard.
It is verified that Map<String, Object?> does not break standard json serialization.

With Object you get some minimal static analysis, with dynamic you can invoke whatever field you want.

There are lints that restrict dynamic and we use them.

polina-c:
> > Can you remove changes in packages/genui_dartantic?
> 
> I needed those changes to get the checks to pass.

Thanks for explanation.

I tried to resolve conflicts on PR by just reverting the changes and it resulted in failures. Sorry.

Can you please resolve the conflicts in smarter way?

csells:
> Can you please resolve the conflicts in smarter way?

Are you saying that you broke it and you'd like me to fix it? Again? After I already fixed it? : )

polina-c:
> > Can you please resolve the conflicts in smarter way?
> 
> Are you saying that you broke it and you'd like me to fix it? Again? After I already fixed it? : )

ha-ha, let me revert my changes.

polina-c:
Okay, I see that github allows me to resolve (unsuccessfully) merge conflicts on your PR with UI, but does not allow to revert the merge.
I am sorry I tried to save your time on conflict resolution, that resulted in waste of your time. I did it because I thought files in other pacjages are changed by mistake, and they just should be reverted.

This is my draft PR created out of your commit: https://github.com/flutter/genui/pull/632

Yes, tests pass, but there are conflicts.

And I have to say I still do not understand why PR that adds completely independent package to the repo, needs to change something on other packages in order to make tests pass. If tests for other packages are failing on your PR, this means they are failing on the main branch too, it maybe it is problem of randomization, and we want to resolve these failures on separate PR.

I see these options to proceed: 

1. You push forward this PR by reverting my merge and fixing conflicts. If tests on other packages are failing, we need to figure it out separately.
2. You create fresh PR, if it is easier for you, and just reference this discussion in the comment
3. I create my PR, based on your commits, reference this discussion on it, and resolve whatever should be resolved, so that you do not need to spend more time.

Thoughts?

csells:
it was a timing issue. the checks pulled down the latest version of dartantic_ai and its dependencies. some of the deps violated semantic versions, so introduced breaking changes. if you're so inclined, I think the think to do is to fix main first with the lastest version of dartantic_ai that fixes the breaking changes and then I can rebase my PR on the updated main.

polina-c:
Yes, if main is broken, it is P0 to fix main. 

Created empty PR to verify if bots on main are healthy: https://github.com/flutter/genui/pull/634
Yes, it is failing. Thank you for highlighting it with your PR. Started working on fix.

And yes, I am inclined, because we are transitioning from highly agile fast development to stable clear solution, so I do not want to merge anything that look strange for me :).

Thank you for your contribution and for your patience.

polina-c:
Issue: https://github.com/flutter/genui/issues/635

csells:
> Thank you for your contribution and for your patience.

No problem at all. This is fun for me. : )

@polina-c let me know when you've landed the fixes to main and I'll rebased my PR.

polina-c:
Landed! 

Thanks to @andrewkolos for the fast fix!

Minor suggestion about this PR: how about using `callId` instead of `interactionId` to reduce number of terms in the package?

csells:
ok @polina-c . I rebased on main, got everything building again and replaced "interactionId" with "callId" as well as most of the other suggestions I well.

also, the latest dartantic_ai works with the latest mistral if that's of interest.

polina-c:
Amazing! Thank you 🙏

Issue #635, opened 2025-12-22
Bots are failing at main.


polina-c:
Failure: https://github.com/flutter/genui/pull/634

PR #636, opened 2025-12-22
fix(deps): pin dartantic_ai to 2.0.3, remove unused dependency on a2a in verdure/client

andrewkolos:
## Description

Fixes https://github.com/flutter/genui/issues/635.

- Pin dartantic_ai to exact 2.0.3 to fix CI failure caused by incompatibility with anthropic_sdk_dart 0.3.1 and mistralai_dart 0.1.1+1
- Remove unused a2a dev_dependency from verdure (not imported anywhere)

Alternatively, we can upgrade to dartantic_ai ^2.1.1 (current) and override `mistralai_dart` to use 0.1.1 instead of 0.1.1+.

## Pre-launch Checklist

- [x] I signed the [CLA].

[changelog]: ../CHANGELOG.md

### DIFF
--- a/packages/genui_dartantic/pubspec.yaml
+++ b/packages/genui_dartantic/pubspec.yaml
@@ -16,7 +16,7 @@ environment:
   flutter: ">=3.35.7 <4.0.0"
 
 dependencies:
-  dartantic_ai: ">=2.0.3 <2.1.0"
+  dartantic_ai: ">=2.0.3 <2.1.0" # TODO(#637): Pinned due to a breakage in latest (2.1.1) due to a downstream breaking change in mistral_ai 0.1.1+.
   flutter:
     sdk: flutter
   genui: ^0.6.0


packages/genui_dartantic/pubspec.yaml
```diff
@@ -16,7 +16,7 @@ environment:
   flutter: ">=3.35.7 <4.0.0"
 
 dependencies:
-  dartantic_ai: ^2.0.1
+  dartantic_ai: 2.0.3
```
polina-c: Can you create issue 'dartantic_ai' should be unpinned and add TODO to unpin when fixed with referenced issue?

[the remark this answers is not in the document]
andrewkolos: done in [fix(deps): use range constraint for dartantic_ai](https://github.com/flutter/genui/pull/636/commits/262eadbc3c8197c522d6c15d4b66e72cc91389db)

packages/genui_dartantic/pubspec.yaml
```diff
@@ -16,7 +16,7 @@ environment:
   flutter: ">=3.35.7 <4.0.0"
 
 dependencies:
-  dartantic_ai: ^2.0.1
+  dartantic_ai: ">=2.0.3 <2.1.0" # TODO(#637): Pinned due to a breakage in latest (2.1.1) due to a downstream breaking change in mistral_ai 0.1.1+.
```
polina-c, line 19: nit and ok to postpone: Can you put full url here, so that it is easy to copy/paste from any code viewer/editor?

polina-c, line 19: nit and ok to postpone: do you mean 'upstream'?

'downstream' normally cannot break anything, because downstream takes whatever is falling down and cannot influence anything that is upper

csells:
the cause is that the author of the mistral etc provider libraries on which dartantic_ai depends is not sticking with semver versioning, so every point release is a breaking change and I gotta panic update dartantic as a result. I have a new update pending right now for his most recent release.

davidmigloz:
Sorry about that! I'll be more careful 🙏

PR #641, opened 2025-12-23
Run CI every hour.

polina-c:


.github/workflows/flutter_packages.yaml
```diff
@@ -23,6 +23,8 @@ on:
       - "packages/**"
       - "examples/**"
       - "tool/**"
+  schedule:
+    - cron: '0 * * * *' # hourly
```
yjbanov, line 29: Consider leaving a comment explaining the purpose behind the hourly runs. It may not be immediately clear why past results against the same SHA are invalid over time. It usually mean that we have some non-hermetic properties in our CI, such as shifting dependency resolution, or inherent non-determinism (e.g. due to LLM temperature).

polina-c: Added, thanks.

No LLM involved so far. If/when we add it, it will be separate job.

Issue #607, opened 2025-12-16
Create separate package genai_primitives


polina-c:
Public discussion: [Lossy Abstraction for Generated Responses](https://github.com/flutter/genui/discussions/599#top)

Internal proposal: go/dart-ai-primitives

- [x] https://github.com/flutter/genui/issues/626
- [x] cleanup / review the package
- [x] update genui to take dependency on the package with local path
- [x] publish the package 
- [x] update genui to take dependency on the package in pub.dev

csells:
I don't think the name "ai_primitives" fits here -- AI is far more general purpose than what we're building. we're building out a set of primitives for use with generative AI, so perhaps "genai_primitives" or "llm_primitives" makes more sense.

csells:
## Types to Pull In from `dartantic_interface`

### **Message/Content Types**

1. **Message Parts** (for multimodal content):
   - `Part` (base class)
   - `TextPart`
   - `DataPart` (for images/binary data)
   - `LinkPart`
   - `ToolPart` (for tool calls and results)
   - `ToolPartKind` (enum: call, result)
2. **Message Container**:
   - `ChatMessage`
   - `ChatMessageRole` (enum: system, user, model)

### **Tool Definition**

1. **`ToolDefinition<TInput>`** (new - extracted from current `Tool<TInput>`):
   - `name`, `description`, `inputSchema`
   - NO execution logic

PR #644, opened 2025-12-24
Updates to make the library applicable for wider range of use cases, and more stable. 

polina-c:

1. Remove ChatMessageRole: this library should not enforce the set of roles, as it can vary from case to case. The role should be part of envelope in client's code.
1. Rename ChatMessage to Message, because this library can be used in chats and in other systems.
1. Use constants for repeating literals 
1. Remove postfix 'Part' from part type literals, to reduce size of messages.
1. Extend test coverage for primitives.
1. Add test for example.
1. Follow Gemini's advises below
1. Replace `dynamic` with `Object?`
1. Refactor `Part` to make the class extendable.
2. Update equality to match Flutter codestyle and to use package `collection` for deep collection comparison

I plan to do more rounds of review before publishing.


polina-c:
@csells , how does it look?

packages/genai_primitives/example/main.dart
```diff
@@ -8,8 +8,24 @@ import 'dart:typed_data';
...
+
+  ChatMessage({required this.role, required this.content});
+
+  ChatMessage.system(Message content)
+    : this(role: Role.system, content: content);
+
+  ChatMessage.user(Message content) : this(role: Role.user, content: content);
+
+  ChatMessage.model(Message content) : this(role: Role.model, content: content);
+}
```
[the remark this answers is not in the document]
polina-c: thanks

packages/genai_primitives/lib/src/message_parts.dart
```diff
@@ -14,52 +14,76 @@ import 'package:path/path.dart' as p;
...
+          return ToolPart.result(
+            callId: content[_Json.id] as String,
+            toolName: content[_Json.name] as String,
+            result: content[_Json.result],
+          );
+        }
+      }(),
+      _ => throw UnimplementedError('Unknown part type: $type'),
+    };
+  }
```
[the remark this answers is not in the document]
polina-c: thanks

packages/genai_primitives/test/genai_primitives_test.dart
```diff
@@ -157,6 +251,32 @@ void main() {
...
+          );
+          expect(part1.argumentsRaw, equals('{}'));
+
+          const part2 = ToolPart.call(
+            callId: 'c2',
+            toolName: 't2',
+            arguments: {'a': 1},
+          );
+          expect(part2.argumentsRaw, equals('{"a":1}'));
+        });
```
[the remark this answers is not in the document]
polina-c: thanks

polina-c:
Options for name for the class ChatMessage:
1. Message (I used it for now, but I am open)
2. MessageBody
3. MessageContent
4. Content

csells:
I think you left one option:

0. ChatMessage 

Can I ask what problem is solved by renaming it?

polina-c:
> I think you left one option:
> 
> 0. ChatMessage
> 
> Can I ask what problem is solved by renaming it?

Sure!

This library is intended to serve very wide range of client applications, interacting with AI.

'Chat' is implying type of interaction, where history is preserved, and UI looks as timeline. 

But, there can be many other types. And, I am sure we do not know these types as they do not exist yet.

Does it help?

csells:
That sounds abstract. My understanding of the goal of this package was to concretely provide base types for chat history, which would eliminate the need to provide mapping layers between GenUI, dartantic and other packages like the Flutter AI Toolkit. 

What concrete scenarios do you have in mind?

polina-c:
> That sounds abstract. My understanding of the goal of this package was to concretely provide base types for chat history, which would eliminate the need to provide mapping layers between GenUI, dartantic and other packages like the Flutter AI Toolkit.
> 
> What concrete scenarios do you have in mind?

For example, interaction with some history-less agent without chat-like UI.
Like agent that is doing google search, without history, using google-search-tool and then selecting responses that are more educational than offering concrete products and services.

csells:
To which agent are you referring that provides such a history-based API without it being conversational? I'm unfamiliar with any that work like that.

packages/genai_primitives/lib/src/message.dart
```diff
@@ -0,0 +1,85 @@
...
+import 'utils.dart';
+
+final class _Json {
+  static const parts = 'parts';
+  static const metadata = 'metadata';
+}
+
+/// A message in a conversation between a user and a model.
+@immutable
+class Message {
```
yjbanov: Is this class intended to be implemented? If not, let's mark it using the `base` modifier.

packages/genai_primitives/lib/src/message.dart
```diff
@@ -0,0 +1,85 @@
...
+
+import 'message_parts.dart';
+import 'utils.dart';
+
+final class _Json {
+  static const parts = 'parts';
+  static const metadata = 'metadata';
+}
+
+/// A message in a conversation between a user and a model.
```
yjbanov: It's not super-clear what "between" means. Does it mean that a `Message` can be sent by the user to the model, and by the model to the user? If so, consider spelling it out, if not in the intro paragraph, then in a second paragraph.

packages/genai_primitives/lib/src/message.dart
```diff
@@ -0,0 +1,85 @@
...
+@immutable
+class Message {
+  /// Creates a new message.
```
yjbanov: Document the arguments. Special cases, such as when parts and metadata are both empty, could be particularly important.

packages/genai_primitives/lib/src/message.dart
```diff
@@ -0,0 +1,85 @@
...
+  const Message.fromParts({this.parts = const [], this.metadata = const {}});
+
+  /// Creates a new message, taking a text string as separate parameter.
+  Message(
+    String text, {
```
yjbanov: Looks like this is a convenience constructor. Typically, the default constructor is the main constructor, and named constructors are conveniences. That is, I'd make `fromParts` the default, and this one named `Message.fromText`.

packages/genai_primitives/lib/src/message.dart
```diff
@@ -0,0 +1,85 @@
...
+    List<Part> parts = const [],
```
yjbanov: What happens if one of the parts is a `TextPart`? Is it OK to have two text parts?

packages/genai_primitives/lib/src/message.dart
```diff
@@ -0,0 +1,85 @@
...
+    Map<String, dynamic> metadata = const {},
+  }) : this.fromParts(parts: [TextPart(text), ...parts], metadata: metadata);
+
+  /// Creates a message from a JSON-compatible map.
```
yjbanov: We should document, or link to the documentation of, the required JSON structure.

packages/genai_primitives/lib/src/message.dart
```diff
@@ -0,0 +1,85 @@
...
+@immutable
+class Message {
+  /// Creates a new message.
+  const Message.fromParts({this.parts = const [], this.metadata = const {}});
+
+  /// Creates a new message, taking a text string as separate parameter.
+  Message(
+    String text, {
+    List<Part> parts = const [],
+    Map<String, dynamic> metadata = const {},
```
yjbanov: Here and throughout the library, avoid using `dynamic` (we should probably enable the `avoid_dynamic` lint too). This is because `dynamic` allows dynamic method invocations, such as `someDynamic.nonExistentMethod()`, which is error prone and negatively affects code tree-shaking (for example, `someDynamic.toString()` could end up retaining all `toString()` implementations globally).

Instead, we should use `Object?`, which is equivalent to `dynamic` type-wise, but prevents dynamic invocations. So we can express JSON as `Map<String, Object?>`, `List<Object?>`, etc.

packages/genai_primitives/lib/src/message.dart
```diff
@@ -0,0 +1,85 @@
+// Copyright 2025 The Flutter Authors.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:meta/meta.dart';
+
+import 'message_parts.dart';
+import 'utils.dart';
+
+final class _Json {
```
yjbanov: Keeping the JSON field names here, separately from the JSON parsing logic (`fromJson`), scatters the code related to the JSON representation of the objects we're working on. I don't know if this will work, but one option might be to use [extension types](https://dart.dev/language/extension-types). Example:

```dart
typedef JsonObject = Map<String, Object?>;

extension type MessageJson(JsonObject json) {
  List<JsonObject> get parts =>
      (json['parts'] as List<Object?>).cast<JsonObject>().toList();
  JsonObject get metadata => json['metadata'] as JsonObject;
}
```

packages/genai_primitives/lib/src/message.dart
```diff
@@ -0,0 +1,85 @@
...
+    parts: (json[_Json.parts] as List<dynamic>)
+        .map((p) => Part.fromJson(p as Map<String, dynamic>))
+        .toList(),
+    metadata: (json[_Json.metadata] as Map<String, dynamic>?) ?? const {},
+  );
+
+  /// The content parts of the message.
+  final List<Part> parts;
+
+  /// Optional metadata associated with this message.
```
yjbanov: Add one empty line between paragraphs.

packages/genai_primitives/lib/src/message.dart
```diff
@@ -0,0 +1,85 @@
...
+  /// Can include information like suppressed content, warnings, etc.
```
yjbanov: grammar nit: start with "This can include..."

packages/genai_primitives/lib/src/message.dart
```diff
@@ -0,0 +1,85 @@
...
+  static const metadata = 'metadata';
+}
+
+/// A message in a conversation between a user and a model.
+@immutable
+class Message {
+  /// Creates a new message.
+  const Message.fromParts({this.parts = const [], this.metadata = const {}});
+
+  /// Creates a new message, taking a text string as separate parameter.
```
yjbanov: It's not clear what "taking a text string" means. What we want to convey is that this constructor creates a message with a `TextPart` created from a text string passed via the `text` argument.

packages/genai_primitives/lib/src/message.dart
```diff
@@ -0,0 +1,85 @@
...
+  /// Converts the message to a JSON-compatible map.
+  Map<String, dynamic> toJson() => {
+    _Json.parts: parts.map((p) => p.toJson()).toList(),
+    _Json.metadata: metadata,
+  };
+
+  @override
+  bool operator ==(Object other) {
+    if (identical(this, other)) return true;
+    return other is Message &&
```
yjbanov: `other is Message` may be incorrect, depending on whether this class is meant to be sub-typed. Consider an extension, such as `ImageMessage`, which adds an `image` field. In the implementation of `ImageMessage.operator==` there's an expectation that `other` has an `image` field and that it is equal to `this.image`. However, this implementation is unaware of the `image` field, and therefore does not check for equality. This means that `imageMsg == baseMsg` can be `false`, while `baseMsg == imageMsg` is `true`.

One way to deal with it might be to do `identical(other.runtimeType, runtimeType)` instead of `is Message`. But I think this all depends on the intent of this class w.r.t. sub-typing. Maybe we don't want it to be sub-typed at all? Should we make it `final`?

packages/genai_primitives/lib/src/message_parts.dart
```diff
@@ -14,52 +14,82 @@ import 'package:path/path.dart' as p;
...
+}
+
+final class _Part {
+  static const text = 'Text';
+  static const data = 'Data';
+  static const link = 'Link';
+  static const tool = 'Tool';
+}
+
 /// Base class for message content parts.
```
yjbanov, line 35: If it's truly a "base class", should we make it formally an `abstract base class`?

packages/genai_primitives/lib/src/message_parts.dart
```diff
@@ -14,52 +14,82 @@ import 'package:path/path.dart' as p;
...
-          result: content['result'],
-        );
-      }
-    }(),
-    _ => throw UnimplementedError('Unknown part type: ${json['type']}'),
-  };
+  factory Part.fromJson(Map<String, dynamic> json) {
+    final Object? type = json[_Json.type];
+
+    switch (type) {
```
yjbanov: If this library defines all the subtypes and automatically resolves to them, I think the idiomatic way to express it in Dart is using [sealed classes](https://dart.dev/language/class-modifiers#sealed).

packages/genai_primitives/lib/src/utils.dart
```diff
@@ -20,3 +20,12 @@ bool mapEquals<K, V>(Map<K, V>? a, Map<K, V>? b) {
   }
   return true;
 }
+
+int mapHashCode(Map<dynamic, dynamic>? map) {
+  if (map == null) return 0;
+  var hash = 0;
```
yjbanov: Let's avoid inventing our own hashing algorithms. I think we can use https://api.dart.dev/dart-core/Object/hashAll.html in this case.

packages/genai_primitives/lib/src/utils.dart
```diff
@@ -20,3 +20,12 @@ bool mapEquals<K, V>(Map<K, V>? a, Map<K, V>? b) {
   }
   return true;
 }
+
+int mapHashCode(Map<dynamic, dynamic>? map) {
+  if (map == null) return 0;
```
yjbanov: Let's avoid redefining the hash code for null. `null` already has hashCode, and it is equal to 2011 on the Dart VM, and 0 on the web. Let's just `return null.hashCode` and let the language/compiler do its thing.

packages/genai_primitives/lib/src/message_parts.dart
```diff
@@ -14,52 +14,82 @@ import 'package:path/path.dart' as p;
...
+  static const data = 'Data';
+  static const link = 'Link';
+  static const tool = 'Tool';
+}
+
 /// Base class for message content parts.
 @immutable
 abstract class Part {
   /// Creates a new part.
   const Part();
```
yjbanov, line 53: Why do we need a constructor for creating an empty part?

packages/genai_primitives/lib/src/message.dart
```diff
@@ -0,0 +1,85 @@
...
+
+  /// The content parts of the message.
+  final List<Part> parts;
+
+  /// Optional metadata associated with this message.
+  /// Can include information like suppressed content, warnings, etc.
+  final Map<String, dynamic> metadata;
+
+  /// Gets the text content of the message by concatenating all text parts.
+  String get text => parts.whereType<TextPart>().map((p) => p.text).join();
```
yjbanov: If I have parts "happy", "new", "year", "world", will the concatenated text be "happynewyearworld"?

packages/genai_primitives/lib/src/message.dart
```diff
@@ -0,0 +1,85 @@
...
+
+  /// Checks if this message contains any tool calls.
```
yjbanov: "Checks" is an unusual way to describe a getter. It's as if it fails if there are no tool calls. Typically we say, "Returns whether this message contains...", or "Returns true if this message contains...".

This comment applies to other getters in the library.

packages/genai_primitives/lib/src/message.dart
```diff
@@ -0,0 +1,85 @@
...
+  bool get hasToolResults =>
+      parts.whereType<ToolPart>().any((p) => p.kind == ToolPartKind.result);
+
+  /// Gets all tool results in this message.
+  List<ToolPart> get toolResults => parts
+      .whereType<ToolPart>()
+      .where((p) => p.kind == ToolPartKind.result)
+      .toList();
+
+  /// Converts the message to a JSON-compatible map.
```
yjbanov: nit: consider "Serializes" instead of "Converts". The word "convert" implies that this same object transforms/mutates to something else, but what this method does is create a new object that represents this object in JSON format.

packages/genai_primitives/lib/src/message.dart
csells: yes, but luckily the text parts include whitespace, so we're good.

polina-c:
> To which agent are you referring that provides such a history-based API without it being conversational? I'm unfamiliar with any that work like that.

(yes, happy New Year!)

I said above not 'history-based' but 'history-less', but it is not very important.

As I said before, we want to cover wide range of stories, published and not published, past and future. It does not matter if such agent exist or no. The landscape of AI-related tools and patterns and practices changes every week. If a story is theoretically possible, and it may benefit from this library, I do not want naming to create confusion. 

And yes, as you noticed, it sounds abstract, and it should be abstract, to allow as many concrete implementations, as possible.

A good side effect of this renaming is that now the name 'ChatMessage' is used in example, and it is good that this name is available for client applications to incapsulate whatever envelope information is needed.

Did I miss something?

csells:
> sounds like you're architecting for an unknown future. that's not something I'm a fan of in general. have you considered `dynamic`? : )

Yes, I am cutting off some opinions of this library that have big chances to be different in different use cases. 

In particular I am separating message body and message envelope: the body should stay in the library (because body has the same structure in most cases), while envelope should be defined in the client application (because content of envelope has big chances to differ from case to case, and thus there is no point to have it in library).  

Does it help?

packages/genai_primitives/lib/src/message.dart
polina-c: yes, added documentation

packages/genai_primitives/lib/src/message.dart
polina-c: Replaced 'creates' with 'Serializes'.

packages/genai_primitives/lib/src/message.dart
polina-c: The goal of the class _Json is to avoid duplication of string literals to make code less error-prone.

Extension type does not help here, but makes the code less reliable by switching from strongly typed members to JSON members.

There are many ways to define enum-like collections of string literals.
I use the one that we already chosen in [other discussion](https://github.com/flutter/genui/issues/358).

packages/genai_primitives/lib/src/message.dart
polina-c: Yes, these text parts are created to support streaming from LLM. And, LLM will produce the parts with spaces and paragraph marks:

"Happy", " new", " year,", " world!\n"

packages/genai_primitives/lib/src/message_parts.dart
polina-c: Because we need parent constructor for all descendants, but implicit constructor is not constant. It is good to redefine it to make constant.

packages/genai_primitives/lib/src/message.dart
yjbanov: Yeah, if there's no clear way to improve it, let's keep it as is.

packages/genai_primitives/lib/src/message.dart
polina-c: Consumed this boilerplate here and in other places: https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#common-boilerplates-for-operator--and-hashcode

packages/genai_primitives/lib/src/message.dart
polina-c: Made it `final` to avoid issues with backward compatibility. If we ever change our mind, it will be easy to remove it or replace `final` with `base`.

packages/genai_primitives/lib/src/message_parts.dart
polina-c: done

thanks

packages/genai_primitives/lib/src/message_parts.dart
polina-c: Nope, we do not want to limit set of parts. Made `Part` extendable and supported the extendability in `Part.fromJson`.

packages/genai_primitives/lib/src/message.dart
```diff
@@ -0,0 +1,102 @@
...
+}
+
+/// A message between participants of the interaction.
+@immutable
+final class Message {
+  /// Creates a new message.
+  ///
+  /// If `parts` or `metadata` is not provided, an empty collections are used.
+  ///
+  /// If there is no parts of type [TextPart], the [text] property
```
yjbanov: grammar: "are no parts"

packages/genai_primitives/lib/src/message.dart
```diff
@@ -0,0 +1,102 @@
...
+  /// will be empty.
+  ///
+  /// If there are many parts of type [TextPart], the [text] property
```
yjbanov: "many" sounds ambiguous. Let's say "more than one".

packages/genai_primitives/lib/src/message.dart
```diff
@@ -0,0 +1,102 @@
...
+  /// will be a concatenation of all of them.
+  /// Many text parts is convenient to have to support
```
yjbanov: Separate paragraphs with one empty line.

grammar: "convenient to have to support" doesn't sound right. Suggestion: "Use multiple text parts to support..."

packages/genai_primitives/lib/src/message.dart
```diff
@@ -0,0 +1,102 @@
...
+  /// streaming of the message.
```
yjbanov: What does "streaming of the message" mean? Standard Dart `Stream`s are asynchronous constructs, but parts are provided as one synchronous `List`.

packages/genai_primitives/lib/src/message.dart
```diff
@@ -0,0 +1,102 @@
...
+  const Message({this.parts = const [], this.metadata = const {}});
+
+  /// Creates text message.
```
yjbanov: grammar: "Creates a text message"

packages/genai_primitives/lib/src/message.dart
yjbanov: The documentation is on a different constructor. This constructor should explain what happens if one passes both `text`, and one or more `TextPart`s.

packages/genai_primitives/lib/src/message.dart
```diff
@@ -0,0 +1,102 @@
...
+  ///
+  /// Converts [text] to a [TextPart] and puts it as a single member of
+  /// the [parts] list.
+  Message.fromText(
+    String text, {
+    List<Part> parts = const [],
+    Map<String, Object?> metadata = const {},
+  }) : this(parts: [TextPart(text), ...parts], metadata: metadata);
+
+  /// Deserializes a message seriealized with [toJson].
```
yjbanov: typo: "serialized"

I'm not sure "serialized with [toJson]" is accurate. This method is _compatible_ with `toJson`, but the original JSON doesn't have to come from `toJson`. It could come from the server.

packages/genai_primitives/lib/src/message.dart
```diff
@@ -0,0 +1,102 @@
...
+  factory Message.fromJson(Map<String, Object?> json) => Message(
+    parts: (json[_Json.parts] as List<Object?>)
+        .map((p) => const PartConverter().convert(p as Map<String, Object?>))
+        .toList(),
+    metadata: (json[_Json.metadata] as Map<String, Object?>?) ?? const {},
+  );
+
+  /// Serializes the message.
```
yjbanov: nit: "Serializes this message to JSON".

packages/genai_primitives/lib/src/message.dart
```diff
@@ -0,0 +1,102 @@
...
+      .whereType<ToolPart>()
+      .where((p) => p.kind == ToolPartKind.result)
+      .toList();
+
+  @override
+  bool operator ==(Object other) {
+    if (identical(this, other)) return true;
+    if (other.runtimeType != runtimeType) return false;
+
+    final deepEquality = const DeepCollectionEquality();
```
yjbanov: nit: `const deepEquality = DeepCollectionEquality();`

packages/genai_primitives/lib/src/message_parts.dart
```diff
@@ -5,139 +5,109 @@
...
-          'id': p.callId,
-          'name': p.toolName,
-          if (p.arguments != null) 'arguments': p.arguments,
-          if (p.result != null) 'result': p.result,
-        };
-        break;
+  @override
+  Part convert(
+    Map<String, Object?> input, {
+    JsonToPartConverter? customConverter,
```
yjbanov: I'm confused by this API. If I pass a `customConverter`, the presumably I already know the type of the `input`. In that case, why do I need to call this method at all? Why wouldn't I call `customConverter.convert(input)` directly?

A common pattern for supporting extensible sets of serializable types is to have a converter registry which is injected into places that need to perform the conversions. Those places do not know the types of objects they are converting. They just trust that the registry will provide a sufficient set of converters.

Here's a primitive converter registry, designed as a global static map:

```dart
/// Converter registry.
///
/// The key of a map entry is the part type. The value is the converter that knows
/// how to convert that part type. The map is prepopulated with standard converters.
/// Custom converters can be added by the developer.
final registry = <String, JsonToPartConverter>{
  _Part.text: TextPartConverter(),
  _Part.data: DataPartConverter(),
};

void main() {
  registry['heading'] = HeadingPartConverter();
}

Part someCodeThatNeedsToConvert(Map<String, Object?> input) {
  // Doesn't need to know anything about supported part types.
  return registry[input['type']]!.convert(input);
}
```

packages/genai_primitives/lib/src/message_parts.dart
```diff
@@ -5,139 +5,109 @@
...
-        content = {
-          'id': p.callId,
-          'name': p.toolName,
-          if (p.arguments != null) 'arguments': p.arguments,
-          if (p.result != null) 'result': p.result,
-        };
-        break;
+  @override
+  Part convert(
+    Map<String, Object?> input, {
```
yjbanov: nit: let's give `input` a proper name (all arguments are "inputs" 😄 ). Perhaps `partJson`?

packages/genai_primitives/lib/src/message_parts.dart
```diff
@@ -5,139 +5,109 @@
...
+  static const text = 'Text';
+  static const data = 'Data';
+  static const link = 'Link';
+  static const tool = 'Tool';
+}
 
 /// Base class for message content parts.
 @immutable
-abstract class Part {
+abstract base class Part {
```
yjbanov, line 51: With a little more understanding of how this works, I think this needs substantially more documentation. In particular, if we want to support custom parts, we should document how to properly create them. The docs should explain about `operator==`, `hashCode`, and the `JsonToPartConverter`. The developer would want to know that all these things need to be covered in order to build a robust custom part implementation.

packages/genai_primitives/lib/src/message_parts.dart
```diff
@@ -5,139 +5,109 @@
...
+  }) {
+    return const PartConverter().convert(
+      json,
+      customConverter: customConverter,
+    );
+  }
 
-  /// The default MIME type for binary data.
-  static const defaultMimeType = 'application/octet-stream';
+  /// Converts the part to a JSON-compatible map.
```
yjbanov: The map needs more than JSON compatibility. The returned value must be a JSON object, containing certain keys, such as "type". Let's document that.

packages/genai_primitives/lib/src/message_parts.dart
```diff
@@ -5,139 +5,109 @@
...
+        return customConverter.convert(input);
     }
-    return {'type': typeName, 'content': content};
   }
 }
 
 /// A text part of a message.
 @immutable
-class TextPart extends Part {
+base class TextPart extends Part {
```
yjbanov: It might be safer to make this `final`. Inheritance complicates equality, hash codes, and serialization. For example, it would be too easy to attempt to call `super.toJson()` from a subclass, and forget to overwrite the part "type". So a `HeadingTextPart` may accidentally end up getting deserialized as a plain `TextPart`.

There's not much code to reuse anyway. The developer could just copy this code and make the necessary changes.

This comment applies to other concrete Part classes below.

pubspec.lock
```diff
@@ -125,10 +125,10 @@ packages:
     dependency: transitive
     description:
       name: characters
-      sha256: f71061c654a3380576a52b451dd5532377954cf9dbd272a78fc8479606670803
+      sha256: faf38497bda5ead2a8c7615f4f7939df04333478bf32e4173fcb06d428b5716b
```
yjbanov, line 120: Did you need to update `pubspec.lock` to make your code changes to work? Or was this by accident?

yjbanov:
@csells @polina-c 

My 2c on the `ChatMessage` => `Message` renaming. In the genUI context some messages coming from the LLM are UI updates/mutations to an existing genUI surface. It would be counterintuitive to think of such messages as "chat messages". So I think the renaming is the right choice. An application that uses this library for a chat application could wrap the `Message` object into a `ChatMessage` object that would contain additional chat-specific semantics. I can even imagine a "chat message" being _one of_ or a _composition of_ a text message, genui surface, and a data message (e.g. containing a generated image).

pubspec.lock
polina-c: I did not mean, but it is unavoidable side effect of storing pubspec.lock in git - sometimes it changes.

polina-c:
@csells , @yjbanov 

We discussed the subject on the doc and agreed:

1. Fixed set of roles will make the class ChatMessage robust and well-reusable between packages.
2. The name Message will stay reserved for the case when we see that class ChatMessage does not fit some stories.
3. The class Parts will make it possible to use parts independently.

See details in go/message-in-ai-primitives

Will update the PR shortly.

packages/genai_primitives/lib/src/message.dart
polina-c: done

packages/genai_primitives/lib/src/message.dart
polina-c: When LLM sends message, it adds parts one by one and, as result, text is split. Stream is not needed here, because  the message is created when full message is received. I removed this comment, because, yes, it causes questions.

packages/genai_primitives/lib/src/message_parts.dart
polina-c: > If I pass a customConverter, the presumably I already know the type of the input. 

That's not true. You have array of parts to deserialize and you do not know the type of each item. 

But I like idea of registry. Thanks.
And, I am surprised Gemini recommended me something that is not common pattern.

PR #646, opened 2025-12-28
[examples] Verdure client cleanup and dependency updates

parlough:


examples/verdure/client/lib/features/screens/upload_photo_screen.dart
```diff
@@ -52,7 +52,9 @@ class UploadPhotoScreen extends ConsumerWidget {
               ),
               const SizedBox(height: 16),
               Text(
-                '''Upload a photo of your front or back yard, and our designers will use it to create a custom vision. Get ready to see the potential.''',
+                'Upload a photo of your front or back yard, '
```
gspencergoog, line 55: I was using triple quotes here because agents seem to have a hard time wrapping these multi-line strings, and triple quotes avoid the lint warning.  I'm fine with converting them back, but it does mean more manual work if the LLMs modify the strings, since they will inevitably fail to wrap them properly.

examples/verdure/client/lib/features/state/loading_state.dart
```diff
@@ -36,7 +36,7 @@ class LoadingState {
       } else if (_isProcessingValue && !isProcessing.value) {
         // Went from true to false, reset messages after a short delay
         // to allow the fade-out animation to complete.
-        Future.delayed(const Duration(milliseconds: 500), clearMessages);
+        Future<void>.delayed(const Duration(milliseconds: 500), clearMessages);
```
gspencergoog, line 39: Perhaps we should enable strict-raw-types (https://dart.dev/tools/analysis#enabling-additional-type-checks)? Or is that too much?

examples/verdure/client/lib/features/screens/upload_photo_screen.dart
parlough: Ah, thanks for the context here and thanks for raising this.

I think since this is an example, we should still try to be idiomatic even if it involves a little extra work.

I do wonder what we can do to make agents better at breaking strings in general. I find while agents don't handle wrapping/splitting strings well, LLM-based tab/auto-complete handles it quite well across different editors. Perhaps similar examples in the edited file are ending up closer in the context window?

examples/verdure/client/lib/features/state/loading_state.dart
parlough: I'm all for it, but since we use a workspace-wide analysis config, I can do that in a follow-up PR for the repository.

I enabled it temporarily for verdure and fixed the one other instance of it.

examples/verdure/client/lib/features/screens/upload_photo_screen.dart
gspencergoog: Works for me.  It think it's just that they are bad at counting characters. I think they need an edit tool that takes a line range and wraps it, contextually, based on language and context in the code.

gspencergoog (approved):
32384589-a60f0e74-c078-11e7-9bc1-e5b5287aea9d

parlough:
Thanks for the review @gspencergoog! I don't seem to have write/merge access on this repository, so feel free to merge when you're happy :D

gspencergoog:
Sorry, I missed that step. :-)  Done!

Issue #628, opened 2025-12-20
Get Points to 160/160 up 10 points from present


hunterino:
Following the convention defined for pub.dev, we should rename the examples/simple_chat to example, and rename the README.md to example.md to conform to the points requirement.  we should also add a examples.md in the examples directory, to give more insight. The points analysis identified several elements that are not dartdoc'd, so the documentation should be updated.

PR #679, opened 2026-01-21
Improve package:genui pub score.

ditman:
# Description

Added some dart documentation and an `example` directory to improve `package:genui` pub score.

* Fixes https://github.com/flutter/genui/issues/628
* Closes https://github.com/flutter/genui/pull/629, addressing @jacobsimionato's comments there.

## Pre-launch Checklist

- [x] I signed the [CLA].

[changelog]: ../CHANGELOG.md

### DIFF
--- a/packages/genui/lib/src/model/a2ui_schemas.dart
+++ b/packages/genui/lib/src/model/a2ui_schemas.dart
@@ -9,10 +9,7 @@ import 'tools.dart';
 
 /// Provides a set of pre-defined, reusable schema objects for common
 /// A2UI patterns, simplifying the creation of CatalogItem definitions.
-class A2uiSchemas {
-  /// This class is not meant to be instantiated.
-  A2uiSchemas._();
-
+abstract final class A2uiSchemas {
   /// Schema for a value that can be either a literal string or a
   /// data-bound path to a string in the DataModel. If both path and
   /// literal are provided, the value at the path will be initialized


ditman:
Apologies @hunterino, I didn't have permissions to push to your original PR, so I had to create a new one. I'm doing this to preserve the authorship of your commits as much as possible!

packages/genui/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - **Fix**: Improved error handling for catalog example loading to include context about the invalid item (#653).
 - **BREAKING**: Renamed `ChatMessageWidget` to `ChatMessageView` and `InternalMessageWidget` to `InternalMessageView` (#661).
 - **Fix**: Pass the correct `catalogId` in `DebugCatalogView` widget (#676).
+- Added some dart documentation and an `example` directory to improve `package:genui` pub score.
```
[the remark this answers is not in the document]
ditman: Your suggestion doesn't accurately describe the change that is being made.

examples/README.md
```diff
@@ -0,0 +1,155 @@
...
+verdure/
+├── server/         # Python A2A server
+│   └── verdure/    # Agent implementation
+└── client/         # Flutter client app
+```
+
+**Run Server:**
+```bash
+cd examples/verdure/server/verdure
+echo "GEMINI_API_KEY=YOUR_API_KEY" > .env
```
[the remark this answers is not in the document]
ditman: This clarification is good. Thanks!

packages/genui/lib/src/model/a2ui_schemas.dart
```diff
@@ -10,6 +10,9 @@ import 'tools.dart';
 /// Provides a set of pre-defined, reusable schema objects for common
 /// A2UI patterns, simplifying the creation of CatalogItem definitions.
 class A2uiSchemas {
+  /// This class is not meant to be instantiated.
+  A2uiSchemas._();
```
ditman: Do we want this private constructor? @jacobsimionato

jacobsimionato: SGTM!

examples/README.md
```diff
@@ -0,0 +1,157 @@
...
+## catalog_gallery
+
+A developer tool for visualizing and testing the core widget catalog. Displays all available `CoreCatalogItems` widgets and allows interaction testing.
+
+**Key Features:**
+- Browse all core catalog widgets
+- Interactive widget testing with event logging
+- Sample file loading support
+
+**Run:**
```
jacobsimionato: It seems like some of the samples already have a sample_app/README.md file. How about we just put a high level summary of what each sample is in this file, and keep the detailed running instructions etc in each specific sample readme. That we we reduce duplication and maintenance cost.

So let's keep the description above here so that people can understand what samples are available quickly, but then move this part about how to run the app to catalog_gallery/README.md to be consistent with the other samples.

I think some of the samples (e.g. travel_app, simple_chat) probably already have running instructions in their app-specific readme.

ditman: Ah yeah, sounds good! I was maintaining @hunterino's original file pretty much unchanged, but I can explode it into each example as needed. Thanks!

packages/genui/lib/src/model/a2ui_schemas.dart
parlough: Another option is to make the class `abstract final` to indicate and enforce it's not meant to be instantiated or inherited from.

examples/README.md
ditman: Done!

packages/genui/lib/src/model/a2ui_schemas.dart
ditman: I like `abstract final` better than the private constructor, thanks @parlough!

ditman:
PTAL @jacobsimionato, I moved content from the big `examples/README.md` to each separate app, and left the global one as the index.

So after this lands, pub.dev for `package:genui` will render a README that tells you to find the actual README in the `examples` directory from the repo, and that one will point you to each specific example.

I'd love the one from pub to just be the same as the one in `examples` (but I don't want to mess with symlinks, which I'm not sure would work!)

jacobsimionato (approved):
Awesome!

PR #680, opened 2026-01-22
Differentiate Part(sealed) and BasePart(extendable).

polina-c:

Sealed class is needed to support dartantic_ai case, that defines common denominator across multiple LLM providers. 

Parts of client's systems that do not need guaranteed compatibility with LLMs can use BasePart.

I am open for better naming.


csells (approved):
lgtm -- loved the sealed parts.

PR #682, opened 2026-01-22
Remove `package:intl` dependency

mosuem:
# Description

Removing `package:intl` for non-internationalized formatting and parsing.

## Pre-launch Checklist

- [ ] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [x] I signed the [CLA].
- [ ] I read the [Contributors Guide].

[changelog]: ../CHANGELOG.md

### DIFF
--- a/packages/json_schema_builder/lib/src/formats.dart
+++ b/packages/json_schema_builder/lib/src/formats.dart
@@ -15,9 +15,10 @@ typedef FormatValidator = bool Function(String);
 ///
 /// Note: the field `Duration` is not supported.
 final Map<String, FormatValidator> formatValidators = {
-  'date-time': (value) => DateTime.tryParse(value) != null,
-  'date': (value) => DateTime.tryParse('${value}T00:00:00Z') != null,
-  'time': (value) => DateTime.tryParse('0000-01-01T$value') != null,
+  'date-time': (value) =>
+      DateTime.tryParse(value) != null && !_isDate(value) && !_isTime(value),
+  'date': _isDate,
+  'time': _isTime,
   'email': EmailValidator.validate,
   'ipv4': (value) {
     final List<String> parts = value.split('.');
@@ -36,3 +37,7 @@ final Map<String, FormatValidator> formatValidators = {
     }
   },
 };
+
+bool _isTime(String value) => DateTime.tryParse('0000-01-01T$value') != null;
+
+bool _isDate(String value) => DateTime.tryParse('${value}T00:00:00Z') != null;
--- a/packages/json_schema_builder/test/schema_test.dart
+++ b/packages/json_schema_builder/test/schema_test.dart
@@ -173,6 +173,12 @@ void main() {
         [],
         strictFormat: true,
       );
+      expectFailuresMatch(schema, '2025-07-29', [
+        ValidationErrorType.formatInvalid,
+      ], strictFormat: true);
+      expectFailuresMatch(schema, '12:34:56Z', [
+        ValidationErrorType.formatInvalid,
+      ], strictFormat: true);
       expectFailuresMatch(schema, 'not-a-date-time', [
         ValidationErrorType.formatInvalid,
       ], strictFormat: true);


mosuem:
I don't quite understand the CI changes though - @gspencergoog any ideas?

packages/json_schema_builder/lib/src/formats.dart
```diff
@@ -13,12 +12,12 @@ typedef FormatValidator = bool Function(String);
...
 /// This is used to validate string formats like 'date-time', 'email', etc.
+///
+/// Note: the field `Duration` is not supported.
 final Map<String, FormatValidator> formatValidators = {
   'date-time': (value) => DateTime.tryParse(value) != null,
-  'date': (value) =>
-      FixedDateTimeFormatter('YYYY-MM-DD').tryDecode(value) != null,
-  'time': (value) =>
-      FixedDateTimeFormatter('hh:mm:ss').tryDecode(value) != null,
+  'date': (value) => DateTime.tryParse(value) != null,
```
gspencergoog: Isn't this a different thing?  For instance, if the string includes only a time, won't this succeed at parsing?

Shouldn't this do something similar to the "time" version, and include a midnight time?

```dart
  'date': (value) => DateTime.tryParse('${value}T00:00:00Z') != null,
```

packages/json_schema_builder/test/schema_test.dart
```diff
@@ -173,11 +173,29 @@ void main() {
...
         strictFormat: true,
       );
+      expectFailuresMatch(schema, 'not-a-date-time', [
+        ValidationErrorType.formatInvalid,
+      ], strictFormat: true);
+    });
+
+    test('format: date', () {
+      final schema = StringSchema(format: 'date');
+      expectFailuresMatch(schema, '2025-07-29', [], strictFormat: true);
```
gspencergoog, line 189: Please add a test that checks for parsing a non-date that looks like a time (e.g. "12:34:56")

packages/json_schema_builder/test/schema_test.dart
```diff
@@ -173,11 +173,29 @@ void main() {
...
         ValidationErrorType.formatInvalid,
       ], strictFormat: true);
     });
 
+    test('format: time', () {
+      final schema = StringSchema(format: 'time');
+      expectFailuresMatch(schema, '12:34:56Z', [], strictFormat: true);
+      expectFailuresMatch(schema, '12:34:56.123Z', [], strictFormat: true);
+      expectFailuresMatch(schema, '12:34:56+01:00', [], strictFormat: true);
+      expectFailuresMatch(schema, 'not-a-time', [
```
gspencergoog, line 206: Please add a test that checks for parsing a non-time that looks like a date: "1234-12-31"

packages/json_schema_builder/lib/src/formats.dart
```diff
@@ -13,31 +12,12 @@ typedef FormatValidator = bool Function(String);
...
-    try {
-      DateFormat('HH:mm:ss').parseStrict(value);
-      return true;
-    } catch (e) {
-      return false;
-    }
-  },
+  'date-time': (value) => DateTime.tryParse(value) != null,
+  'date': (value) => DateTime.tryParse(value) != null,
+  'time': (value) => DateTime.tryParse('0000-01-01T$value') != null,
```
gspencergoog: Won't this now allow a "Z" where it didn't used to?

For instance, before it would have failed to parse the time "12:34:56Z", but now it will succeed.

gspencergoog:
The CI issues might not be a problem with this PR.  It looks like someone messed with the submodule specification or CI setup.  I'll investigate.

gspencergoog:
The CI issue was just a temporary GitHub outage.

packages/json_schema_builder/lib/src/formats.dart
mosuem: That's a good idea!

Technically we would need an RFC3339 compliant parser, but this should be good enough.

packages/json_schema_builder/lib/src/formats.dart
mosuem: The behavior before was kind of undefined and depended on the locale's interpretation of the `HH:mm:ss` skeleton. I think this should be more precise. https://datatracker.ietf.org/doc/html/rfc3339#section-5.6 allows for a Z after the time, if I am not mistaken.

mosuem:
Thanks for the LGTM! I don't have merge rights here. so feel free to merge.

PR #683, opened 2026-01-22
Rename Part to StandardPart and BasePart to Part.

polina-c:


polina-c:
> What's the motivation for this rename? `Part` is what has been used in most implementations I've seen, and it's pretty self explanatory, and matches what LLMs are using in their APIs.

Thanks for asking.

In this library BasePart is extendable (users can define their own parts for their systems), and StandardPart is sealed as something that most LLM providers support and that is de-facto industry standard.

If we name it Part, it will trigger question 'how it is different from BasePart?' and necessity to read documentation.

Does it help?

(I am open for other names suggestion)

gspencergoog:
> If we name it Part, it will trigger question 'how it is different from BasePart?' and necessity to read documentation.

Works for me.  Should  `BasePart` be called `Part` then?  I feel like adding "base" to the name is an anti-pattern (what if later we add a superclass?)

polina-c:
> > If we name it Part, it will trigger question 'how it is different from BasePart?' and necessity to read documentation.
> 
> Works for me. Should `BasePart` be called `Part` then? I feel like adding "base" to the name is an anti-pattern (what if later we add a superclass?)

Yes, I like it. Thank you for suggesting!

PR #690, opened 2026-02-01
Use null-aware elements per latest lint update

kevmoo:


kevmoo:
@gspencergoog ?

gspencergoog:
I'm fine with this, as long as it works with Flutter Stable version of Dart (which I think it does).

gspencergoog (approved):

PR #707, opened 2026-02-10
Turn on dependabot auto-merge

gspencergoog:
# Description

This converts the dependabot checks for github actions to autosubmit. This just means that it will submit once approved and all tests pass.

### DIFF
--- a/examples/verdure/server/uv.lock
+++ b/examples/verdure/server/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 2
+revision = 3
 requires-python = ">=3.13"
 resolution-markers = [
     "python_full_version >= '3.14'",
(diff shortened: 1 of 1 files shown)


.github/dependabot.yaml
```diff
@@ -9,6 +9,8 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    labels:
+      - "autosubmit"
```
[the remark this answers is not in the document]
gspencergoog: This is fine, because it still requires approval before it will submit.  It will just auto-merge once the tests pass and the PR is approved.

examples/verdure/server/uv.lock
jacobsimionato, line 1: I think this artifact-product-foundry is the airlocked version of the python package repo that is difficult to use outside corp? Can we avoid it in this open source repo? I never quite figured out how to set this up properly for the A2UI repo, seems like a pain.

gspencergoog: Ugh.  Yeah, it's not supposed to be there.  We should be able to avoid it... somehow.  I'll look and find out.  Alternatively, we could just let dependabot update it. I think that's what I'll do for this PR, and just revert the uv.lock changes.

Issue #716, opened 2026-02-13
[0.9 branch] system instruction construction process is quite manual


yjbanov:
**Describe the bug**

Right now you have to write code like this:

```
    final systemInstruction =
        '''You are a helpful assistant who chats with a user.
Your responses should contain acknowledgment of the user message.

IMPORTANT: When you generate UI in a response, you MUST always create
a new surface with a unique `surfaceId`. Do NOT reuse or update
existing `surfaceId`s. Each UI response must be in its own new surface.

<a2ui_schema>
$a2uiSchema
</a2ui_schema>

${StandardCatalogEmbed.standardCatalogRules}

${PromptFragments.basicChat}''';
```

Which, unless you poke around sample apps, you wouldn't get right. Also, if we change how the schema is intended to be embedded into the SI, the SI code you wrote before will break with the new version of A2UI.

**Expected behavior**

I think I'd like (if possible) two utilities:

1. An A2UI SI constructor that keeps the construction process on rails. When I upgrade the version of the GenUI SDK, I will also automatically get the new and improved constructor.
2. If my app is introducing agentic functionality for the first time, I won't have an existing AI pipeline to plug my GenUI pipeline into. It would be useful to have a prompt builder utility that helps me build a complete SI from scratch using some ergonomic Dart API (e.g. a builder pattern).

PR #719, opened 2026-02-13
Create prompt_builder.dart

polina-c:
Fixes https://github.com/flutter/genui/issues/716

### DIFF
--- a/docs/Design_Proposal.md
+++ b/docs/Design_Proposal.md
@@ -245,7 +245,7 @@ controller.handleMessage(
 
 ```
 Surface(
-  genUiContext: myHost.contextFor('main-surface'),
+  surfaceContext: myHost.contextFor('main-surface'),
 )
 ```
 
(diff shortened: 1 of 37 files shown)


packages/genui/lib/src/facade/prompt_builder.dart
```diff
@@ -0,0 +1,53 @@
...
+
+  late final String prompt = () {
+    final String a2uiSchema = A2uiMessage.a2uiMessageSchema(
+      catalog,
+    ).toJson(indent: '  ');
+
+    return '''
+${instructions ?? ''}
+
+IMPORTANT: When you generate UI in a response, you MUST always create
```
gspencergoog, line 34: This isn't always the case, it's only for the chat use case.

polina-c: Thanks.

Renamed constructor to be chat-only.

PR #722, opened 2026-02-17
Small fixes.

polina-c:
1. Update docs to run with api key. 

2. Add spacing to column and row:

## Before:
Screenshot 2026-02-16 at 9 23 05 PM

## After:

Screenshot 2026-02-16 at 9 25 44 PM

### DIFF
--- a/packages/genui/lib/src/catalog/basic_catalog_widgets/column.dart
+++ b/packages/genui/lib/src/catalog/basic_catalog_widgets/column.dart
@@ -9,8 +9,9 @@ import '../../model/a2ui_schemas.dart';
 import '../../model/catalog_item.dart';
 import '../../model/ui_models.dart';
 import '../../primitives/simple_items.dart';
-import 'ui_styling.dart';
 import 'widget_helpers.dart';
+
+const _verticalColumnSpacing = 8.0;
 
 final _schema = S.object(
   properties: {
@@ -117,7 +118,7 @@ final column = CatalogItem(
           mainAxisAlignment: _parseMainAxisAlignment(columnData.justify),
           crossAxisAlignment: _parseCrossAxisAlignment(columnData.align),
           mainAxisSize: MainAxisSize.min,
-          spacing: UiStyling.verticalColumnSpacing,
+          spacing: _verticalColumnSpacing,
           children: childIds.map((componentId) {
             final explicitWeight =
                 getComponent(componentId)?.properties['weight'] as int?;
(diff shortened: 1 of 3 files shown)


packages/genui/lib/src/catalog/basic_catalog_widgets/shared_constants.dart
```diff
@@ -0,0 +1,3 @@
+class SharedConstants {
+  static const double spacing = 8;
```
gspencergoog: I think "spacing" is too generic of a term.  I think it should be "verticalSpacing" and "horizontalSpacing" at least, because they often want to be different values.  You might even go as far as "horizontalRowSpacing" and "verticalColumnSpacing", although I feel less strongly about that, since it ties it to one component.

polina-c: I see two options: 

1. Have constant for each case 
2. Have just one constant 'spacing' for all types of spacing that are needed and will be needed

As this constant is not confugurable at the moment and we do not know if it will be configurable, it maybe (2) is more efficient, just to enable some spacing for places where zero spacing looks weird.

(1) looks as overdesign for me with current design and it is not clear if it will change.

For the note: DevTools have just three constants for spacing: spacing, tightSpacing and looseSpacing (or something like this).

Thoughts?

polina-c: Added comment for the constant.

gspencergoog: I already would like to set the horizontal spacing differently from vertical, and it should probably eventually be based on the screen size and use case.

For instance, in Material Design, 16dp is the default gutter spacing between columns (in a Row) on mobile, but it's 24dp on desktop.  And depending on the use case, spacing between rows (in a Column) should have 8, 16, or 24dp spacing.  So, I'd suggest that for now the default for Columns should be 8dp, but Rows should be 16dp.

I understand that this is just a stopgap to help things look better for now, but we should probably think more about the details of this, because it's not as simple as a single number.

polina-c: I see. Updated.

polina-c: And renamed the class as it now have different meaning.

polina-c: As agreed in chat, placed the constants at the top of the file where they are used.

PR #744, opened 2026-02-21
Improve doc organization.

polina-c:


gspencergoog:
I think the assets folder should stay within `docs`.  They are assets just for the docs, not something that needs to be at the top level (nothing else will use them), and it means all the links have to climb out of the directory to find the assets (e.g. "../assets/foo.png").

The 	`design.md` file is not really a specification, it's just detailed documentation of the design, which is why I put it into the docs folder.  An agent will still see it in the docs folder, so it should have the same effect for them.

I actually question that the specs folder should be called that, since the things in it aren't specifications, really.  It seems more like information for contributors.

polina-c:
> I think the assets folder should stay within `docs`. They are assets just for the docs, not something that needs to be at the top level (nothing else will use them), and it means all the links have to climb out of the directory to find the assets (e.g. "../assets/foo.png").
> 
> The `design.md` file is not really a specification, it's just detailed documentation of the design, which is why I put it into the docs folder. An agent will still see it in the docs folder, so it should have the same effect for them.
> 
> I actually question that the specs folder should be called that, since the things in it aren't specifications, really. It seems more like information for contributors.

Yes! Let's find structure that we both like.

I would like to see clear separation:
1. Documents for users (human or AI), 'docs' in this PR
2. Documents for contributors (human or AI), 'specs' in this PR

We also have `assets` that contain assets for both folders above.

This is another option to structure the docs:

- docs
    - usage
    - contributing
    - assets

Thoughts?

gspencergoog:
> This is another option to structure the docs:
> 
> - docs
>     - usage
>     - contributing
>     - assets

Yes, let's put it all under docs.  That sounds like a good organization.

polina-c:
Thanks, applied.

gspencergoog:
Can we move specs into the docs folder too?  Then all the docs will be in one place.

polina-c:
/gemini review

Verify links to assets are not broken and all assets are used.

polina-c:
/gemini

> These launch configurations for `custom_backend` and `tic_tac_toe (dartantic)` have been removed. If these examples are no longer relevant or maintained, their removal is appropriate for cleaning up the development environment. However, if these are still valid examples that should be runnable, they should be updated to reflect the current project structure or kept if they are still functional.

Can you verify the examples do not exist any more?

.gemini/GEMINI.md
```diff
@@ -1,5 +1,5 @@
 # Gemini Code Assistant Context
 
-Follow the specifications in `specs/README.md`.
+Follow the specifications in `docs/contributing/README.md`.
```
gspencergoog:
```suggestion
Follow the guidelines in `docs/contributing/README.md`.
```

Issue #735, opened 2026-02-19
Autopopulate component name from `CatalogItem` name


redbrogdon:
**Is your feature request related to a problem? Please describe.**
In one of my apps, I define a CatalogItem called WorkoutCard, including the name as a parameter to the constructor:

```dart
final workoutCard = CatalogItem(
  name: 'WorkoutCard',
  dataSchema: workoutCardSchema,
  widgetBuilder: (itemContext) {
    ...
  },
);
```

I also define a schema for the agent to use when creating instances of it, which includes the name as well:

```dart
final workoutCardSchema = S.object(
  properties: {
    'component': S.string(enumValues: ['WorkoutCard']),
    'title': S.string(description: 'The title of the workout'),
    ...
  },
  required: ['title', 'exercises'],
);
```

Up until ten minutes ago, I hadn't realized that the `component` field in the schema properties was necessary. I'd been leaving it out, and most of the time Gemini was smart enough to pick the right catalog item ~90% of the time. No error was triggered when I left it out, nor was the package able to autopopulate that field with the name I'd given my catalog item. 90% is great, but I'd like for Gemini to get it right 100% of the time, and catching or fixing my mistake sooner would have helped.

**Describe the solution you'd like**
Ideally the package would either update the schema with the name given in the constructor for `CatalogItem` or include an assert to let me know that I've failed to include the component name when defining the schema in the first place.

PR #757, opened 2026-02-24
feat(catalog): consolidate component name into CatalogItem constructor

gspencergoog:
# Description

Extracts the `name` property directly to the `CatalogItem` constructor. The `CatalogItem` now explicitly takes `name` as a required parameter string, and automatically injects the required `component` discriminator into the data schema when the `dataSchema` is evaluated.

This reduces repetitive boilerplate across the 19 basic catalog widgets by removing the redundant `component` schema definition mapping and the manual inclusion of `component` in the properties list. The constructor parameter accepts the `Schema` type allowing the developer to instantiate `S.object(…)` smoothly while structurally enforcing `ObjectSchema` externally.

Fixes https://github.com/flutter/genui/issues/735

### DIFF
--- a/packages/genui/lib/src/model/catalog_item.dart
+++ b/packages/genui/lib/src/model/catalog_item.dart
@@ -103,8 +103,10 @@ final class CatalogItem {
 
   /// The schema definition for this widget's data.
   ///
-  /// Automatically injects the `component` discriminator property required by
-  /// A2UI.
+  /// It should contain all of the component specific properties, but not the
+  /// `component` discriminator property, which will be automatically injected
+  /// using the [name]. If the `component` property is already defined in the
+  /// schema, it will be ignored.
   ObjectSchema get dataSchema {
     final Map<String, Object?> originalMap = _originalSchema.value;
     final Map<String, Object?> properties =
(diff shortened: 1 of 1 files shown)


packages/genui/lib/src/model/catalog_item.dart
```diff
@@ -90,17 +90,41 @@ final class CatalogItem {
   /// Creates a new [CatalogItem].
   const CatalogItem({
     required this.name,
-    required this.dataSchema,
+    required Schema dataSchema,
```
jacobsimionato, line 93: Add some documentation for this parameter explaining that it should include all the component-specific properties of the component, but not its name.

jacobsimionato (approved):
Nice!!

packages/genui/lib/src/model/catalog_item.dart
```diff
@@ -90,17 +90,41 @@ final class CatalogItem {
...
     required this.widgetBuilder,
     this.exampleData = const [],
     this.isImplicitlyFlexible = false,
-  });
+  }) : _originalSchema = dataSchema;
 
   /// The widget type name used in JSON, e.g., 'TextChatMessage'.
   final String name;
 
+  final Schema _originalSchema;
```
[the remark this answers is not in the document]
gspencergoog: This won't work because `Schema.object` (and so `S.object`) is a factory constructor aliased to the `ObjectSchema` constructor, and so it would result in a type mismatch, because at the end, the factory returns type `Schema`.

packages/genui/lib/src/model/catalog_item.dart
gspencergoog: Done.

Issue #751, opened 2026-02-22
Implement prompt builder.


polina-c:
Implement prompt builder as designed in go/genui-what-is-allowed.

PR #760, opened 2026-02-24
Add `copyWith` and `concatenate` methods to `ChatMessage`

polina-c:
Contributes to https://github.com/flutter/genui/issues/751

### DIFF
--- a/packages/genai_primitives/lib/src/chat_message.dart
+++ b/packages/genai_primitives/lib/src/chat_message.dart
@@ -186,10 +186,8 @@ final class ChatMessage {
       );
     }
 
-    return ChatMessage(
-      role: role,
+    return copyWith(
       parts: [...parts, ...other.parts],
-      metadata: metadata,
       finishStatus: finishStatus ?? other.finishStatus,
     );
   }


packages/genai_primitives/lib/src/chat_message.dart
```diff
@@ -162,6 +162,27 @@ final class ChatMessage {
...
+  /// Concatenates two messages.
+  ///
+  /// The role of the first message is used; the second message's role
+  /// is ignored.
+  ///
+  /// The finish status of the first message is used if it's not null;
+  /// otherwise, the second message's finish status is used.
+  ///
+  /// If metadata keys are the same, the first message's metadata is used.
+  ChatMessage operator +(ChatMessage other) => ChatMessage(
```
gspencergoog: I feel like adding operators often gets into a place where it's unclear what the return type is, or what it actually means.  Is there a clear use case for this operator?  How does it handle adding of other types (e.g. can I add a `String` to a `ChatMessage`)?

I'd be inclined to skip adding this unless it really unlocked a _lot_ of developer simplicity and code readability.

In this particular implementation, I feel like the fact that the operation isn't commutative (it is order dependent) makes it less accurate to use "+" as the operator, since addition is commutative.

packages/genai_primitives/lib/src/chat_message.dart
```diff
@@ -175,6 +196,19 @@ final class ChatMessage {
         finishStatus == other.finishStatus;
   }
 
+  /// Creates a copy of this message with optional fields replaced.
+  ChatMessage copyWith({
```
gspencergoog, line 209: This seems like a good addition.

packages/genai_primitives/lib/src/chat_message.dart
polina-c: I agree. Thanks for catching this.

I was inspired by operator '+' for List in Dart ()that is NOT commutative), but now I see that, yes, it is questionable.

Removed '+' and improved 'concatenate'.

I need 'concatenate' for prompt operations in promptBuilder.

One of features of promptBuilder is to combine user message with system prompt. And, potentially, both may contain non-text parts, that makes the class ChatMessage very veluable.

How does it sound?

gspencergoog: Much better!

packages/genai_primitives/lib/src/chat_message.dart
```diff
@@ -162,6 +162,53 @@ final class ChatMessage {
...
+  ///   are not null.
+  /// - Throws [ArgumentError] if metadata keys conflict and values
+  ///   are different.
+  ///
+  /// Otherwise:
+  /// - The role of the first message is used; the second message's role
+  ///   is ignored.
+  /// - The finish status of the first message is used if it's not null;
+  /// otherwise, the second message's finish status is used.
+  /// - If metadata keys are the same, the first message's metadata is used.
```
gspencergoog: If they are the same, it doesn't matter which one is used.  Do you mean "not the same"?

packages/genai_primitives/lib/src/chat_message.dart
```diff
@@ -162,6 +162,53 @@ final class ChatMessage {
...
+        throw ArgumentError(
+          'There are metadata keys that conflict: $conflictingKeys',
+        );
+      }
+    }
+
+    return ChatMessage(
+      role: role,
+      parts: [...parts, ...other.parts],
+      metadata: {...other.metadata, ...metadata},
```
gspencergoog: This is a little weird, in that for the other keys, `other` is primary, but for this one it is overwritten if there already is a `metadata` with that item.  Also, this might get weird if two incompatible metadata items with different keys existed. E.g. if the mimetype of this was set to "image/jpg" and had some image EXIF metadata too, and if `other` was "text/plain": the resulting metadata would contain EXIF data for the image, but the mime type is now plain text.

packages/genai_primitives/lib/src/chat_message.dart
```diff
@@ -162,6 +162,53 @@ final class ChatMessage {
...
+      finishStatus: finishStatus ?? other.finishStatus,
```
gspencergoog, line 191: Shouldn't the finish status reflect the "worst" status?  For example, if `this` has a "stopped" finish status, but `other` has a failure, we don't want to erase the failure.

packages/genai_primitives/lib/src/chat_message.dart
```diff
@@ -162,6 +162,53 @@ final class ChatMessage {
...
+
+      if (conflictingKeys.isNotEmpty) {
+        throw ArgumentError(
+          'There are metadata keys that conflict: $conflictingKeys',
+        );
+      }
+    }
+
+    return ChatMessage(
+      role: role,
```
gspencergoog: In the non-throwing case, this seems problematic to override: You could end up adding user text parts to the system message, for instance.

packages/genai_primitives/lib/src/chat_message.dart
```diff
@@ -162,6 +162,53 @@ final class ChatMessage {
...
+  /// - Throws [ArgumentError] if metadata keys conflict and values
+  ///   are different.
+  ///
+  /// Otherwise:
+  /// - The role of the first message is used; the second message's role
+  ///   is ignored.
+  /// - The finish status of the first message is used if it's not null;
+  /// otherwise, the second message's finish status is used.
+  /// - If metadata keys are the same, the first message's metadata is used.
+  ChatMessage concatenate(ChatMessage other, {bool throwOnConflicts = true}) {
```
gspencergoog: I wonder if it would be better to just call this `concatenateParts` and just supply it with `List<Part>` to avoid the complications of combining the other fields of `ChatMessage` with this message.

Or maybe just remove `throwOnConflicts` entirely and make it always throw of these things don't combine well.  The non-throwing version just feels like it will cause more problems than it will help.

packages/genai_primitives/lib/src/chat_message.dart
```diff
@@ -162,6 +162,53 @@ final class ChatMessage {
...
+        throw ArgumentError('Roles must match for concatenation');
+      }
+
+      if (finishStatus != null &&
+          other.finishStatus != null &&
+          finishStatus != other.finishStatus) {
+        throw ArgumentError('Finish statuses must match for concatenation');
+      }
+
+      final Iterable<String> conflictingKeys = metadata.keys
```
gspencergoog: It's not just that you want to see if the values of existing keys are different, but that the resulting combination of metadata keys is still valid.  So there's more validation that would be needed here, and I'm not sure there's enough information to know which metadata keys are combinable (or even what the set of possible values would be).

It might be necessary to require that the metadata be deeply identical?

polina-c: done

packages/genai_primitives/lib/src/chat_message.dart
```diff
@@ -189,7 +189,7 @@ final class ChatMessage {
     return ChatMessage(
```
gspencergoog: This could use `copyWith` and only copy parts and status.

packages/genai_primitives/lib/src/chat_message.dart
polina-c: it is always throwing now, if conflicts

packages/genai_primitives/lib/src/chat_message.dart
polina-c: it is always throwing now, if conflicts

packages/genai_primitives/lib/src/chat_message.dart
polina-c: if both are not null, it fails now

packages/genai_primitives/lib/src/chat_message.dart
polina-c: it is always throwing now, if conflicts

packages/genai_primitives/lib/src/chat_message.dart
polina-c: yes, removed throwOnConflicts, thanks for good suggestion

gspencergoog (approved):

PR #769, opened 2026-02-25
Update the "iteration task" issue template to "sprint task" 

jacobsimionato:
This should make it easier to create tasks in the sprint.

- Add "sprint ready" tag, to automate adding to the project.
- Add default P2 priority to skip triage

### DIFF
--- a/.github/ISSUE_TEMPLATE/sprint_task.md
+++ b/.github/ISSUE_TEMPLATE/sprint_task.md
@@ -2,7 +2,7 @@
 name: Sprint Task
 about: An issue used sprint planning
 title: ''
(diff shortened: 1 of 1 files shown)


polina-c (approved):

polina-c:
Do we want to add label 'front-line-treaged' instead of P2?

Normally current sprint is P1, so i would just skip priority

jacobsimionato:
> front-line-treag

Good idea. I added both. In the A2UI repo, we just assign a priority to mark as triaged rather than having front-line-handled, so I wanted to bring some consistency between Gen UI and A2UI repos here, by ensuring both have a priority.

PR #777, opened 2026-02-26
Initial implementation of prompt builder.

polina-c:
Contributes to https://github.com/flutter/genui/issues/751

### DIFF
--- a/examples/simple_chat/lib/chat_session.dart
+++ b/examples/simple_chat/lib/chat_session.dart
@@ -22,7 +22,7 @@ class ChatSession extends ChangeNotifier {
     final Catalog catalog = BasicCatalogItems.asCatalog(
       systemPromptFragments: [
         '''
-When you need additional information from user, try to use the component '${BasicCatalogItems.choicePicker.name}' to ask for it.
+When you need additional information from the user, try to use the component '${BasicCatalogItems.choicePicker.name}' to ask for it.
 ''',
         '''
 If there is no way to itemize all the options, either use the component '${BasicCatalogItems.textField.name}' or add option 'Other' to the '${BasicCatalogItems.choicePicker.name}'.


packages/genui/lib/src/facade/prompt_builder.dart
```diff
@@ -15,52 +40,44 @@ class PromptBuilder {
...
+      ...instructions,
+      'Use the provided tools to respond to the user using rich UI elements.',
+      ...catalog.instructions,
+      '''
 <a2ui_schema>
 $a2uiSchema
 </a2ui_schema>
+''',
+      BasicCatalogEmbed.basicCatalogRules,
+    ];
```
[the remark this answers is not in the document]
polina-c: It is addressed in SurfaceOperations and is covered in tests

packages/genui/lib/src/facade/prompt_builder.dart
```diff
@@ -2,65 +2,337 @@
...
+Requires `surfaceId`, `path` and `value`. 
+''',
+  ),
+  deleteSurface(
+    name: 'deleteSurface',
+    explanation: 'Deletes a surface.',
+    properties: '''
+Requires `surfaceId`.
+''',
+  );
```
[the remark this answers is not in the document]
polina-c: we do 'trim' to address it

packages/genui/lib/src/facade/prompt_builder.dart
```diff
@@ -2,65 +2,337 @@
...
+      catalog,
+    ).toJson(indent: '  ');
+
+    final fragments = <String>[
+      ...systemPromptFragments,
+      'Use the provided tools to respond to user using rich UI elements.',
+      ...catalog.systemPromptFragments,
+      allowedOperations.systemPromptFragment,
+      'A2UI Message Schema:\n$a2uiSchema',
+      ?_encodedDataModel(clientDataModel),
```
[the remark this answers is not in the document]
polina-c: nope, it works

packages/genui/lib/src/model/catalog.dart
```diff
@@ -40,6 +47,13 @@ interface class Catalog {
   /// e.g. 'com.example.my_catalog'.
   final String? catalogId;
 
+  /// Instructions for UI generation.
+  ///
+  /// This can include explanation when to use which catalog items.
+  ///
+  /// The items will be incorporated into the system prompt.
+  final List<String> systemPromptFragments;
```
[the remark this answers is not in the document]
polina-c: thank you, fixed and added test coverage

packages/genui/lib/src/facade/prompt_builder.dart
```diff
@@ -2,65 +2,337 @@
...
+  ///
+  /// [prefix] is a prefix to be added to the system prompt.
+  /// Is useful when you want to emphasize the importance of this fragment.
+  static String currentDate({String prefix = ''}) =>
+      'Current Date: '
+      '${DateTime.now().toIso8601String().split('T').first}';
+
+  /// Code execution restriction.
+  ///
+  /// This is useful when AI may need to execute code.
```
gspencergoog: This doesn't seem like an accurate comment.

packages/genui/lib/src/facade/prompt_builder.dart
```diff
@@ -2,65 +2,337 @@
...
+  ///
+  /// [prefix] is a prefix to be added to the system prompt.
```
gspencergoog: Prefix doesn't appear to be used here.

packages/genui/lib/src/facade/prompt_builder.dart
```diff
@@ -2,65 +2,337 @@
...
+    required this.properties,
+  });
+
+  final String name;
+  final String explanation;
+  final String properties;
+
+  String get tickedName => '`$name`';
+
+  static String explainMessages(Set<ProtocolMessages> operations) {
```
gspencergoog, line 176: This should probably be Iterable, not Set.  You can call toSet() internally if you want to ensure uniqueness.

packages/genui/lib/src/facade/prompt_builder.dart
```diff
@@ -2,65 +2,337 @@
...
+
+/// Creates prompt for allowed surface operations.
+final class SurfaceOperations {
+  SurfaceOperations({
+    required this.create,
+    required this.update,
+    required this.delete,
+    required this.dataModel,
+  });
+  SurfaceOperations.createOnly({required bool dataModel})
```
gspencergoog, line 251: This seems overly complicated, and doesn't scale. Imagine adding 10 more message types...  Why not just accept an iterable of `ProtocolMessages`, or even just leave it as the `SurfaceOperations` constructor above, and just make all the bools default to `false`?

packages/genui/lib/src/facade/prompt_builder.dart
```diff
@@ -2,65 +2,337 @@
...
+    if (update) {
+      parts.add('''
+To update an existing UI:
+1. Output an ${ProtocolMessages.updateComponents.tickedName} message with the existing `surfaceId` and the new component definitions.
+''');
+    }
+
+    parts.add('''
+IMPORTANT:
+- Do not use tools or function calls for UI generation. Use JSON text blocks.
```
gspencergoog: This might not be a great piece of advice in all cases.  Some implementations might use functions to generate A2UI. For instance, if you had an MCP wrapper around it.

packages/genui/lib/src/facade/prompt_builder.dart
```diff
@@ -2,65 +2,337 @@
...
+- Ensure all JSON is valid and fenced with ```json ... ```.
```
gspencergoog: We should probably have a better fence than this.  Something like:
<a2ui_json>
\```json
\```
</a2ui_json>

or 
-----BEGIN_A2UI_JSON-----
\```json
\```
-----END_A2UI_JSON-----

There will still be JSON fences around it but we should try and still allow the generation to generate *other* non-A2UI JSON.  Right now that isn't possible.

packages/genui/lib/src/facade/prompt_builder.dart
```diff
@@ -2,65 +2,337 @@
...
+
+**OUTPUT FORMAT:**
+You must output a VALID JSON object representing one of the A2UI message types ($operationsFormatted).
+- Do NOT use function blocks or tool calls for these messages.
```
gspencergoog: Again, not always good advice.

packages/genui/lib/src/facade/prompt_builder.dart
```diff
@@ -2,65 +2,337 @@
...
+- You can treat the A2UI schema as a specification for the JSON you typically output.
```
gspencergoog, line 309: I'd narrow this to only apply to A2UI JSON.  This sounds like the spec should replace all JSON output forms.

packages/genui/lib/src/facade/prompt_builder.dart
```diff
@@ -2,65 +2,337 @@
...
+- You may include a brief conversational explanation before or after the JSON block if it helps the user, but the JSON block must be valid and complete.
+- Ensure your JSON is fenced with ```json and ```.
```
gspencergoog, line 311: Again, better fences needed.  And why is this repeated?

packages/genui/lib/src/facade/prompt_builder.dart
polina-c: replaced with 'This is useful to communicate limitations of code execution to the AI.'

does it look better?

packages/genui/lib/src/facade/prompt_builder.dart
polina-c: It is convenient/ergnomic to prepare this parameter as set:

```
    final operations = <ProtocolMessages>{};
    if (create) {
      operations.addAll([
        ProtocolMessages.createSurface,
        ProtocolMessages.updateComponents,
      ]);
    }
    if (update) {
      operations.add(ProtocolMessages.updateComponents);
    }
    if (delete) {
      operations.add(ProtocolMessages.deleteSurface);
    }
    if (dataModel) {
      operations.add(ProtocolMessages.updateDataModel);
    }
```

It is extra code and mental effort for reader to convert it back and forth.

Am I missing something?

packages/genui/lib/src/facade/prompt_builder.dart
polina-c: Yes, it is good idea to have all bools defaulting to false. Added.

If we have 10 operations, we will not create 10^2 constructors. 

I believe though having some helper methods will add some convenience, readability and ergonomics.

packages/genui/lib/src/facade/prompt_builder.dart
polina-c:
```
-----A2UI_JSON_SCHEMA_START-----
```json
$a2uiSchema
```
-----A2UI_JSON_SCHEMA_END-------
```

File: packages/genui/lib/src/facade/prompt_builder.dart (hunk 9 above)
Comment (polina-c) [reply packages/genui/lib/src/facade/prompt_builder.dart]: defined TechnicalPossibilities

File: packages/genui/lib/src/facade/prompt_builder.dart (hunk 11 above)
Comment (polina-c) [reply packages/genui/lib/src/facade/prompt_builder.dart]: defined TechnicalPossibilities

File: packages/genui/lib/src/facade/prompt_builder.dart (hunk 7 above)
Comment (gspencergoog) [reply packages/genui/lib/src/facade/prompt_builder.dart:176]: A `Set` is an `Iterable`, so you can still do that.  But you can also pass a `List` (which is also `Iterable`), and there's no chance that the passed-in `operations` collection will be accidentally modified inside this function, since iterable can't be modified.

File: packages/genui/lib/src/facade/prompt_builder.dart (hunk 8 above)
Comment (gspencergoog) [reply packages/genui/lib/src/facade/prompt_builder.dart:251]: Well, but is `SurfaceOperations(update: true)` really more ergonomic than `SurfaceOperations.updateOnly()`? They both are pretty clear, and exactly the same length.

And why isn't it `SurfaceOperations.update()` if `SurfaceOperations.createAndUpdate()` exists?  Ahh, because then it clashes with the member, I see.  But then maybe the other should be `SurfaceOperations.createAndUpdateOnly()` to be consistent?

What rules will we use to know what messages to include in the "convenience" set?  Even adding one more common message to the combinations would double the list (8 instead of 4 combinations), plus having `createAndUpdateAndFoobar` gets crazy long...

File: examples/simple_chat/lib/chat_session.dart [hunk 14]
```diff
@@ -19,7 +19,16 @@ class ChatSession extends ChangeNotifier {
     _transport = AiClientTransport(aiClient: aiClient);
 
     // 2. Initialize Catalog & Controller
-    final Catalog catalog = BasicCatalogItems.asCatalog();
+    final Catalog catalog = BasicCatalogItems.asCatalog(
+      systemPromptFragments: [
+        '''
+When you need additional information from user, try to use the component '${BasicCatalogItems.choicePicker.name}' to ask for it.
```
gspencergoog:
```suggestion
When you need additional information from the user, try to use the component '${BasicCatalogItems.choicePicker.name}' to ask for it.
```

polina-c:
> I'm a little concerned that golden files aren't the best way to test this (might be fragile), but we can assess that later and change it if they become annoying.

Goldens are great, because they are easy to fix, with visibility on PRs about what is changed.

Text goldens are better than picture goldens, because they are not flacky.

But, yes, some extra effort. 

Yes, let's see how it goes.

Issue #764, opened 2026-02-25
Eval: evaluate prompt builder configuration against real models in CI.


polina-c:
- [x] [Configure stub key](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-secrets) on github
- [x] Get key - go/eval-for-open-source
- [x] Create smoke test
- [x] Evaluate prompt builder
- [ ] Document important details in go/flutter-genui-internal

polina-c:
'Normal' secrets are not visible for workflows. We need [environment secrets](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-secrets#creating-secrets-for-an-environment).

I need admin permissions to set them up.

@jacobsimionato or @yjbanov can you please give me permissions?

PR #809, opened 2026-03-18
Add test for simple chat.

polina-c:
Fixes https://github.com/flutter/genui/issues/764.

Screenshot 2026-03-17 at 9 03 24 PM

### DIFF
--- a/examples/eval/test/simple_chat_test.dart
+++ b/examples/eval/test/simple_chat_test.dart
@@ -32,17 +32,15 @@ void main() {
       await tester.sendMessageAndWaitForResponse(message);
     }
 
-    final Iterable<String> surfaceIds = tester.surfaceIds();
-
     expect(
-      surfaceIds.length,
+      tester.surfaceIds().length,
       isPositive,
       reason: 'Model should produce surfaces',
     );
 
     tester.verifyEvents();
 
-    tester.reporter.failIfIssuesFound();
+    tester.failIfIssuesFound();
   }, timeout: const Timeout(Duration(seconds: 120)));
 }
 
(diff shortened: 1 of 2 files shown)


examples/eval/test/test_infra/issue_reporter.dart
```diff
@@ -0,0 +1,24 @@
...
+  void expect(bool expectation, String issue) {
+    if (expectation) return;
+    // ignore: avoid_print
+    print('Issue: $issue');
+    _issuesFound++;
+  }
+
+  void failIfIssuesFound() {
+    if (_issuesFound > 0) {
+      fail('Found $_issuesFound issues');
```
jacobsimionato: Would it make sense to actually print all the issues here, in a multi-line string or something?

examples/eval/test/simple_chat_test.dart
```diff
@@ -0,0 +1,130 @@
...
+
+    expect(
+      surfaceIds.length,
+      isPositive,
+      reason: 'Model should produce surfaces',
+    );
+
+    tester.verifyEvents();
+
+    tester.reporter.failIfIssuesFound();
```
jacobsimionato: Nit: Maybe it makes sense to implement failifIssuesFound on tester, so that the developer doesn't need to know about "reporter"?

jacobsimionato (approved):
Awesome!! Our first real integration test!!

examples/eval/test/simple_chat_test.dart
polina-c: good idea!

examples/eval/test/test_infra/issue_reporter.dart
polina-c: Reporter prints them as they appear with prefix 'Issue:', and it looks good in the log. Added note how they are prefixed.

PR #817, opened 2026-03-19
Fix flutter action version

andrewkolos:
Fixes https://github.com/flutter/genui/issues/816.

We could prevent this from happening again by either 1) modifying the org's allow-list to use a wild card for the version if we deem that safe enough or 2) operate like flutter/flutter and instead of using subosito/flutter-action, we just checkout flutter as part of the CI process.

In the meantime, I added code comments that the PR review bot will hopefully pick up on and caution anyone else from making my same mistake.

### DIFF
--- a/examples/eval/test/test_infra/issue_reporter.dart
+++ b/examples/eval/test/test_infra/issue_reporter.dart
@@ -1,3 +1,7 @@
+// Copyright 2025 The Flutter Authors.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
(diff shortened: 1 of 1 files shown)


andrewkolos:
> In the meantime, I added code comments that the PR review bot will hopefully pick up on and caution anyone else from making my same mistake.

The review bot doesn't run on this PRs, so this won't work. Anyway, I'll probably file a tech debt issue and me or someone else can re-evaluate during one of our regularly occurring cleanup weeks.

Issue #825, opened 2026-03-24
Implement pure dart ValueNotifier in flutter/genui.


polina-c:
The implementation should make it possible for Flutter Framework to switch to the new implementation.

PR #827, opened 2026-03-24
Experimental version of pure dart value notifier and what is needed for it.

polina-c:
Contributes to https://github.com/flutter/genui/issues/825

This package is intended for experimentation.

The code is not expected to be of production quality.

See more in [TEMP_README.md](https://github.com/polina-c/genui/blob/6182ce2961225af19dbf5d08c44803823011297c/packages/ui_primitives/TEMP_README.md).

### DIFF
--- a/packages/ui_primitives/lib/ui_primitives.dart
+++ b/packages/ui_primitives/lib/ui_primitives.dart
@@ -2,7 +2,24 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-export 'src/foundation/assertions.dart' show UiError, UiErrorDetails;
+export 'src/foundation/assertions.dart'
+    show
+        ErrorDescription,
+        PartialStackFrame,
+        RepetitiveStackFrameFilter,
+        StackFilter,
+        StackFrame,
+        UiError,
+        UiErrorDetails;
+export 'src/foundation/diagnostics.dart'
+    show
+        DiagnosticLevel,
+        DiagnosticPropertiesBuilder,
+        Diagnosticable,
+        DiagnosticsNode,
+        DiagnosticsSerializationDelegate,
+        DiagnosticsTreeStyle,
+        TextTreeConfiguration;
 export 'src/foundation/listenable.dart' show Listenable, ValueListenable;
 export 'src/foundation/value_notifier.dart' show ValueNotifier;
 export 'src/primitives/basics.dart' show VoidCallback;
--- a/packages/ui_primitives/pubspec.yaml
+++ b/packages/ui_primitives/pubspec.yaml
@@ -5,7 +5,7 @@
 name: ui_primitives
 description: Highly experimental package.
 repository: https://github.com/flutter/genui/tree/main/packages/ui_primitives
-version: 0.0.1-dev-001
+version: 0.0.1-dev-002
 
 resolution: workspace
 


packages/ui_primitives/TEMP_README.md
```diff
@@ -0,0 +1,22 @@
...
+
+This text is not part of 'normal' README.md, because it is not relevant for the package's landing page on pub.dev.
+
+## Changes
+
+The code is copied from flutter and dart:ui
+with the following changes:
+
+1. Word `flutter` is replaced with `ui` in some APIs (for example, `FlutterError` -> `UiError`).
+2. `_lerpDouble` is renamed to `lerpDoubleNonNull`, but remails private.
```
ditman:
```suggestion
2. `_lerpDouble` is renamed to `lerpDoubleNonNull`, but remains private.
```

packages/ui_primitives/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+## 1.0.0
```
ditman, line 1: Doesn't match the one in the pubspec.yaml (are we publishing to pub.dev?)

packages/ui_primitives/test/smoke_test.dart
```diff
@@ -0,0 +1,23 @@
+// Copyright 2025 The Flutter Authors.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:test/test.dart';
+import 'package:ui_primitives/ui_primitives.dart';
+
+void main() {
+  test('smoke', () {
```
ditman: Remember to rename 'smoke' to something descriptive once adding more testing coverage.

packages/ui_primitives/lib/ui_primitives.dart
```diff
@@ -0,0 +1,7 @@
+// Copyright 2025 The Flutter Authors.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+export 'src/foundation/assertions.dart' show UiError, UiErrorDetails;
+export 'src/foundation/listenable.dart';
+export 'src/foundation/value_notifier.dart';
```
ditman: I would recommend to export with `show` even for these files too, to ensure that we don't add anything to the public API of the package by mistake. (Not sure how much this package is intended to grow.)

packages/ui_primitives/lib/src/web.dart
```diff
@@ -0,0 +1,171 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// This code is copied from `package:web` which still needs its own
+// documentation for public members. Since this is a shim that users should not
+// use, we ignore this lint for this file.
+// ignore_for_file: public_member_api_docs
+
+/// A stripped down version of `package:web` to avoid pinning that repo in
+/// Flutter as a dependency.
```
ditman: Can this package depend on `packge:web` instead of copying this file over?

Since this file is not exported from the ui_primitives.dart (I don't think it should!), I suppose flutter will keep a separate copy of this file as well?

In case we need to keep both this copy and flutter's: is this file trimmed down to the minimum amount of copypaste needed by this package?

packages/ui_primitives/lib/src/foundation/value_notifier.dart
```diff
@@ -0,0 +1,427 @@
...
+/// adding items; as an exercise, consider adding buttons to remove items from
+/// the list as well.
+///
+/// ** See code in examples/api/lib/widgets/transitions/listenable_builder.3.dart **
+/// {@end-tool}
+///
+/// See also:
+///
+///  * [ValueNotifier], which is a [_ChangeNotifier] that wraps a single value.
+class _ChangeNotifier implements Listenable {
```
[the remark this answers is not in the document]
ditman: Is there value in exposing a general ChangeNotifier class? Or do we only care about ValueNotifier? (in any case, it can be done later, it's much easier to add a class to the exports of the package than to remove it)

packages/ui_primitives/lib/src/foundation/_platform_io.dart
```diff
@@ -0,0 +1,47 @@
+// Copyright 2025 The Flutter Authors.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:io';
+
+import 'assertions.dart';
+import 'constants.dart';
+import 'platform.dart' as platform;
```
ditman, line 9: The cyclic imports between _platform_x.dart and platform.dart are breaking my mind.

Similarly to package:web below, is there an opportunity to use package:platform here, instead of having all this code? (maybe?)

packages/ui_primitives/lib/src/foundation/basic_types.dart
```diff
@@ -0,0 +1,274 @@
+// Copyright 2025 The Flutter Authors.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// @docImport 'dart:ui';
+library;
+
+import 'dart:collection';
+
+// COMMON SIGNATURES
```
ditman, line 10: Maye consider maybe moving `VoidCallback` here? (or maybe unifying these types in a single location of the package?) It's a little bit weird that `basics.dart` is pretty much empty.

packages/ui_primitives/lib/src/foundation/bitfield.dart
ditman, line 1: Is this "bitfield.dart" being used anywhere in the packge? I can't find where. It is also not being exported from the ui_primitives.dart. Can this be removed and maybe added when it's needed later?

packages/ui_primitives/lib/src/dart_ui/lerp.dart
```diff
@@ -0,0 +1,30 @@
...
+  assert(b.isFinite, 'Cannot interpolate between finite and non-finite values');
+  assert(t.isFinite, 't must be finite when interpolating between values');
+  return a * (1.0 - t) + b * t;
+}
+
+/// Linearly interpolate between two doubles.
+///
+/// Same as [lerpDouble] but specialized for non-null `double` type.
+double lerpDoubleNonNull(double a, double b, double t) {
+  // This doesn't match _lerpInt to preserve specific behaviors when dealing
```
ditman, line 27: I read in the docs that this is a copy of _lerpDouble; does _lerpInt also need to be copied over?

polina-c: _lerpInt is not used, so what's the point?

packages/ui_primitives/lib/src/foundation/basic_types.dart
polina-c: Yes, gowing forward we will need to organize things well.

But for now, while we are experimenting, I prefer structure that separates two things:
- files that are copied almost as is
- separately moved items

packages/ui_primitives/lib/src/foundation/_platform_io.dart
polina-c: Good question.

I prefer to just move this code for now.

Most likely it will be beneficial to move the entire web.dart and re-export it from Flutter Framework.

The cyclic imports are copied as is. I do not like them as well. I plan to resolve the cycles going forward (see TODO in [TEMP_README.md](https://github.com/polina-c/genui/blob/6182ce2961225af19dbf5d08c44803823011297c/packages/ui_primitives/TEMP_README.md)), but not now.

packages/ui_primitives/lib/src/foundation/basic_types.dart
polina-c: At this stage I prefer copied files to stay with minimal changes.

Yes, in future it may be good move.

packages/ui_primitives/CHANGELOG.md
polina-c: Yes, we are publishing dav versions so far. As I understand non-release versions should not be listed here.

packages/ui_primitives/lib/ui_primitives.dart
polina-c: Good suggestion. Applied.

packages/ui_primitives/lib/src/web.dart
polina-c: yes!

good idea

referenced

packages/ui_primitives/lib/src/foundation/bitfield.dart
polina-c: good point!

removed

polina-c:
Answering https://github.com/flutter/genui/pull/827#discussion_r2989835213

ChangeNotifier is not exposed. It is internal. I use it as implementation of ValueNotifier.

PR #831, opened 2026-03-26
Prototype of a2ui_core library

jacobsimionato:


packages/a2ui_core/lib/src/protocol/catalog.dart
```diff
@@ -0,0 +1,39 @@
...
+abstract class FunctionApi {
+  String get name;
+  String get returnType;
+  Schema get argumentSchema;
+}
+
+/// A function implementation that can be registered with a catalog.
+abstract class FunctionImplementation extends FunctionApi {
+  /// Executes the function. Can return a static value or a [ValueListenable].
+  dynamic execute(Map<String, dynamic> args, DataContext context, [CancellationSignal? cancellationSignal]);
```
jacobsimionato, line 22: I think we can do better than returning 'dynamic' here. `Object` at a minimum, or else some custom return type we define that is like a union type of the possible return types.

PR #840, opened 2026-03-30
Make error injectable and cleanup not needed code.

polina-c:
Contributes to https://github.com/flutter/genui/issues/834

### DIFF
--- a/packages/ui_primitives/lib/src/value_notifier.dart
+++ b/packages/ui_primitives/lib/src/value_notifier.dart
@@ -93,7 +93,7 @@ class _ChangeNotifier implements Listenable {
   static bool debugAssertNotDisposed(_ChangeNotifier notifier) {
     assert(() {
       if (notifier._debugDisposed) {
-        throw FrameworkErrorReporter.instance.error(
+        throw FrameworkErrorReporter.instance.createError(
           'A ${notifier.runtimeType} was used after being disposed.\n'
           'Once you have called dispose() on a ${notifier.runtimeType}, it '
           'can no longer be used.',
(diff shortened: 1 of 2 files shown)


packages/ui_primitives/lib/src/value_notifier.dart
```diff
@@ -94,7 +93,7 @@ class _ChangeNotifier implements Listenable {
   static bool debugAssertNotDisposed(_ChangeNotifier notifier) {
     assert(() {
       if (notifier._debugDisposed) {
-        throw UiError(
+        throw FrameworkErrorReporter.instance.error(
```
ditman: Why not make `_FrameworkError` public so you can directly `throw FrameworkError('Reason')` it?

If you want to keep the type private, I think it's a little bit odd having to go through the `.instance` to access the `.error` method. Does `.error` need anything from the instance? If not, the `error` method could be made static to make this throw a little bit less verbose?

ditman: Also, why not use FrameworkErrorReporter.instance.report here? Or a simplified version of `report` that accepts a string? That way everything goes through the `.instance` of the reporter?

packages/ui_primitives/lib/ui_primitives.dart
```diff
@@ -2,60 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-export 'src/foundation/assertions.dart'
```
ditman, line 5: So much deleted stuff! Where did that go?

packages/ui_primitives/lib/src/value_notifier.dart
polina-c: We want the package to be able to throw and we want the stack trace to start in the place of problem, not in the place that created error.

The point of 'instance' is to allow framework to configure their own error reporting. It is needed to keep Flutter error handling backward compatibility.

Flutter will override 'instance' and replace error with FlutterError.

FrameworkError is just default and we want it to be overriden.

polina-c: I will add documentation to make this clear.

packages/ui_primitives/lib/ui_primitives.dart
polina-c: back to Flutter :)

PR #841, opened 2026-03-31
fix: prompt builder misleading LLM to return free text amongst A2UI messages

juan-vgv:
# Description

The current output format instructions in prompt builder are "misleading" the LLM into returning free text in between A2UI messages. When removing this hint the LLM is a lot less likely to return these free text chunks.

There's no reference (that I can find!) in the A2UI spec that hints at servers or clients dealing with free text in the middle of A2UI messages. See for example the [v0.9 example stream](https://a2ui.org/specification/v0.9-a2ui/#example-stream), a "clean" list of JSON messages, no free text in between.

Personally I think that with this in mind `ConversationContentReceived` could / should be removed as a `ConversationEvent`, but that might be taking it a step too far for now. 

## Pre-launch Checklist

- [x] I signed the [CLA].
- [ ] I updated/added relevant documentation (doc comments with `///`).

[changelog]: ../CHANGELOG.md

### DIFF
--- a/packages/genui/lib/src/facade/prompt_builder.dart
+++ b/packages/genui/lib/src/facade/prompt_builder.dart
@@ -311,7 +311,7 @@ To update an existing UI:
       _fenced('''
 When constructing UI, you must output a VALID A2UI JSON object representing one of the A2UI message types ($_operationsFormatted).
 - You can treat the A2UI schema as a specification for the JSON you typically output.
-- You may include a brief conversational explanation before or after the JSON block if it helps the user, but the JSON block must be valid and complete.
+- The JSON block must be valid and complete.
 - Ensure your JSON is fenced with ```json and ```.
 ''', sectionName: 'OUTPUT FORMAT'),
     );
--- a/packages/genui/test/facade/prompt_builder_test.golden/all_operations_with_dataModel_false.txt
+++ b/packages/genui/test/facade/prompt_builder_test.golden/all_operations_with_dataModel_false.txt
@@ -119,7 +119,7 @@ To update an existing UI:
 -----OUTPUT_FORMAT_START-----
 When constructing UI, you must output a VALID A2UI JSON object representing one of the A2UI message types (`createSurface`, `updateComponents`, `deleteSurface`).
 - You can treat the A2UI schema as a specification for the JSON you typically output.
-- You may include a brief conversational explanation before or after the JSON block if it helps the user, but the JSON block must be valid and complete.
+- The JSON block must be valid and complete.
 - Ensure your JSON is fenced with ```json and ```.
 -----OUTPUT_FORMAT_END-----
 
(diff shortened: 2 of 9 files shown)


gspencergoog:
I understand why you might want this change, but if the application wants both text and UI output, making this change requires that the application fire off separate inferences for them. There can be multiple A2UI blocks in the output, and that also facilitates streaming multiple UI updates in the same inference.

I can see removing the "- You may include a brief conversational explanation before or after the JSON block if it helps the user, but the JSON block must be valid and complete." part, so that you're free to add additional instructions to prevent it without contradiction, but by the same token, we shouldn't add "Do not include any conversational text or explanations outside of the JSON blocks." to allow apps that want that.

juan-vgv:
Thanks @gspencergoog!

> if the application wants both text and UI output

I still fail to understand why a GenUI application would like to mix these 2. But if anyone wants to give the LLM the chance to reply with text then they can provide a text catalog item. And then the LLM speaks A2UI and the SDK deals with A2UI.

The GenUI SDK handling both A2UI messages _and_ free text in a single request feels like an advanced use case. And it's currently getting in the way of a "pure" GenUI app. Because when given the chance the LLM does indeed split its response between A2UI and free text messages.

EDIT: 

> Do not include any conversational text or explanations outside of the JSON blocks

That is not in the PR, that's only a Gemini suggestion, but it has not been applied as far as I can see?

gspencergoog:
> That is not in the PR, that's only a Gemini suggestion, but it has not been applied as far as I can see?

Oh, you're right, sorry, I misread things.

> I still fail to understand why a GenUI application would like to mix these 2. But if anyone wants to give the LLM the chance to reply with text then they can provide a text catalog item. And then the LLM speaks A2UI and the SDK deals with A2UI. The GenUI SDK handling both A2UI messages and free text in a single request feels like an advanced use case. 

In a chat context, this is common.  You use A2UI to provide some UI in the chat, and also include plain text in the same inference output.  E.g. a flight status A2UI block followed by "Here's your flight status.  Is there anything else I can help you with?"  Having it all in one inference reduces latency, and the agent doesn't need to think about the chat UI itself.

> And it's currently getting in the way of a "pure" GenUI app. Because when given the chance the LLM does indeed split its response between A2UI and free text messages.

In a "pure" GenUI app, you should only have to add a prohibition to the generated prompt to prevent it, right?

juan-vgv:
@gspencergoog 

> In a "pure" GenUI app, you should only have to add a prohibition to the generated prompt to prevent it, right?

Maybe, can't really say at this point. I'm probably naive but it's proving harder that I would've thought to get LLMs to produce a stream of structured A2UI messages (not even wrapped in ```json blocks) and for the GenUI SDK to consume it.

In any case, this is a step in that direction, the PR should be good to merge?

gspencergoog (approved):

gspencergoog:
> In any case, this is a step in that direction, the PR should be good to merge?

Yep, looks good.

PR #846, opened 2026-04-01
Fix eval tests

gspencergoog:
# Description

This fixes the `packages/genui/examples/eval` tests, since they are failing.  They also aren't being run by CI, presumably because they make actual backend calls.

### DIFF
--- a/examples/eval/test/simple_chat_test.dart
+++ b/examples/eval/test/simple_chat_test.dart
@@ -90,22 +90,48 @@ class _ChatSessionTester {
     var content = 0;
     var waiting = 0;
     final errors = <String>[];
+
+    var currentTurnCreates = 0;
+    var currentTurnUpdates = 0;
+    var turnCount = 0;
+
+    void verifyTurn() {
+      if (turnCount > 0) {
+        reporter.expect(
+          currentTurnCreates <= 1,
+          'Turn $turnCount should create at most 1 surface',
+        );
+        reporter.expect(
+          currentTurnUpdates == currentTurnCreates,
+          'Turn $turnCount should have matching creates ($currentTurnCreates) '
+          'and updates ($currentTurnUpdates)',
+        );
+      }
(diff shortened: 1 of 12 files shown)


examples/eval/test/simple_chat_test.dart
polina-c, line 124: Can you please update doc comment for ConversationWaiting?

When we get it, does it mean AI stopped responding and is waiting for new request/comment?

If yes, in the test we should check that 1 create + 1 update are received before 'waiting'.


And that waiting is received before sending new message

packages/genui/lib/src/engine/surface_controller.dart
```diff
@@ -234,10 +234,6 @@ interface class SurfaceController implements SurfaceHost, A2uiMessageSink {
         final DataModel model = _store.getDataModel(surfaceId);
         model.update(path, value);
```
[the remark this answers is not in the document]
gspencergoog: Data model is already reactive: this causes a second spurious update.

examples/eval/test/simple_chat_test.dart
gspencergoog: > When we get it, does it mean AI stopped responding and is waiting for new request/comment?

No, it actually means the opposite. `ConversationWaiting` is fired when the system starts waiting for the AI response.

The sequence of events during a single turn (started by the user sending a message) actually looks like this:

1. `ConversationWaiting` (Fired immediately when we submit our message to the transport)
2. `ConversationSurfaceAdded` (Whenever the AI starts generating a surface)
3. `ConversationComponentsUpdated` (Whenever the AI streams UI updates)

No event is currently emitted when the AI finishes its turn (though the internal `ConversationState.isWaiting` boolean switches to `false`).

Added more doc comments and updated `simple_chat_test.dart` to verify that exactly one create and one update happen between ConversationWaiting events.

PR #850, opened 2026-04-02
Create Dart replicas for Flutter's notifiers.

polina-c:
Contributes to https://github.com/flutter/genui/issues/826

See go/pure-dart-observability


packages/ui_primitives/lib/src/value_notifier.dart
```diff
@@ -8,51 +8,55 @@ library;
...
+import 'error_reporting.dart';
 import 'primitives.dart';
-import 'private_leak_tracking.dart';
 
-/// A class that can be extended or mixed in that provides a change notification
-/// API using [VoidCallback] for notifications.
-///
-/// It is O(1) for adding listeners and O(N) for removing listeners and
-/// dispatching
-/// notifications (where N is the number of listeners).
```
gspencergoog, line 21: These seem like useful dartdocs.  Why not just update them?

packages/a2ui_core/lib/src/listenable/error_reporting.dart
```diff
@@ -0,0 +1,52 @@
...
+  /// Depending on settings, it may throw an exception, log an error,
+  /// or debug-stop the execution.
+  ///
+  /// In current implementation it just throws an error,
+  /// but it may be reconsidered in future.
+  static void report(ListenableErrorDetails details) =>
+      _reportedErrors.add(details);
+
+  static Iterable<ListenableErrorDetails> get reportedErrors => _reportedErrors;
+  static final _reportedErrors = <ListenableErrorDetails>[];
```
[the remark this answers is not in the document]
polina-c: added comment

gspencergoog (approved):

packages/ui_primitives/lib/src/value_notifier.dart
polina-c: The link I placed instead goes to exactly this dartdoc. 

Reference instead of copying seems to protect more from modifying these classes. They should not be modified, because they are temporary and should be replaced with dash-wide dart classes.

Added this note to the doc comments.

Issue #811, opened 2026-03-18
Unified Architecture: Restructure GenUI to align with A2UI v0.9 (Decoupled Core + Flutter Renderer)


jacobsimionato:
# Design Document: A2UI Flutter Unified Architecture (v0.9)

## 1. Introduction & Motivation

Currently, the Flutter GenUI library processes A2UI streams and manages state using a monolithic, immutable data structure (`SurfaceDefinition`). Whenever an `updateComponents` or `updateDataModel` message arrives, the `SurfaceController` applies the changes and pushes a completely new `SurfaceDefinition`. The `Surface` widget listens to this global change and rebuilds the entire component tree. 

As UI complexity grows, this approach has several drawbacks:
1. **Performance (Lack of Granular Reactivity):** Typing in a `TextField` updates the Data Model, which triggers a full surface rebuild.
2. **Coupling:** A2UI message parsing, JSON Pointer resolution, and Flutter widget building are tightly entangled.
3. **Cross-Platform Inconsistency:** The A2UI Web Renderers (React, Angular, Lit) all share a unified, framework-agnostic state engine (`@a2ui/web_core`). Flutter's implementation is entirely bespoke, making it harder to maintain parity and share catalog logic.

**The Solution:** Implement the **A2UI Unified Architecture**. We will split the library into a **Platform-Agnostic State Engine** (`genui_core`, pure Dart) and a **Framework-Specific Renderer** (`genui`, Flutter).

## 2. The Great Decoupling (Package Structure)

We will introduce a new package into the monorepo:
*   **`packages/genui_core` (Pure Dart):** Has absolutely zero dependency on `package:flutter`. It is responsible for parsing A2UI messages, maintaining the live state tree, resolving JSON pointers, and evaluating logic/expressions.
*   **`packages/genui` (Flutter):** Depends on `genui_core` and `package:flutter`. It contains the Flutter `Surface` widget, the `BasicCatalog` Flutter widgets (Material/Cupertino), and layout delegates.

## 3. Architectural Changes: Old vs. New

This section maps the current Flutter-centric classes to their new, decoupled counterparts.

### 3.1. State Management: From Static to Live Models
Currently, `SurfaceDefinition` acts as a static snapshot of the UI.
*   **Deprecated:** `SurfaceDefinition`
*   **New (`genui_core`):** `SurfaceModel`, `SurfaceComponentsModel`, and `ComponentModel`.
    *   Instead of a static map, `SurfaceComponentsModel` acts as a live registry.
[...]

PR #857, opened 2026-04-06
feat: introduce a2ui_core library

### LINKED ISSUE #828: Create a2ui_core library, independent of genui
Let's implement an a2ui_core library in isolation, following https://github.com/google/A2UI/blob/main/specification/v0_9/docs/renderer_guide.md .

To include:
- Model classes
- MessageProcessor
- Context classes
- Basic catalog API declarations

This way, we can build a Jaspr renderer etc in top of this soon, and refactor the Gen UI SDK to be based on it later.

andrewkolos:
Resolves https://github.com/flutter/genui/issues/828

Builds on #831 toward objective 3.1 as described in #811: establishing a framework-agnostic state engine for the v0.9 A2UI spec (tracking issue: #828). See individual commit messages for adjustments/fixes made on top of #831.

## ⚠️ For reviewers

The diff is large, but every file is interdependent and I didn't see any productive way to split up the change into smaller ones. I tried to match the structure of the [renderer guide](https://github.com/google/A2UI/blob/main/specification/v0_9/docs/renderer_guide.md) in the code, which should make it possible to individually check each module against the corresponding section in the guide.

Resources:
* A2UI v0.9 spec: https://github.com/google/A2UI/blob/main/specification/v0_9/docs/a2ui_protocol.md
* v0.9 renderer guide: https://github.com/google/A2UI/blob/main/specification/v0_9/docs/renderer_guide.md
* The `web_core` implementation, which can be used as a reference to compare this one with: https://github.com/google/A2UI/tree/main/renderers/web_core

## Known gaps left out of scope of this PR

* Type coercion. Note: the renderer guide doesn't specify an ideal place in the pipeline for it to occur.
* Logical functions. Left out of scope for a future phase per #811.
* CancellationSignal is not threaded through function invocation. No applications yet.
* `addLifecycleListener` on `MessageProcessor`. Currently implemented using a different API shape (reading `SurfaceGroupModel` directly). To be re-evaluated.

## Some observations on the renderer guide/a2ui spec (to be considered later)
* ComponentModel.onUpdated is typed as EventSource<ComponentModel>, which causes equality-based reactive systems to silently drop repeated updates
* DataContext constructor takes only a DataModel, but function call resolution requires catalog access
* The DynamicValue resolution algorithm (recursive function call resolution, transitive dependency tracking) is unspecified
* Action event context values must be resolved at dispatch time, not bind time--not documented
* Type coercion table doesn't specify where in the pipeline to apply it

## **Next steps after this is merged**
Per #828,
* Wire `a2ui_core` into `packages/genui`
* Refactor `SurfaceController` to use `MessageProcessor`
* Update Flutter widgets to use `ComponentContext`
* Achieve/test for granular reactivity in the actual Flutter renderer

## Pre-launch Checklist

- [x] I signed the [CLA].

[changelog]: ../CHANGELOG.md


packages/a2ui_core/lib/src/core/contexts.dart
```diff
@@ -0,0 +1,135 @@
...
+        }
+        final Object? result = surface.catalog.invoker(call.call, args, this);
+        if (result is ValueListenable) {
+          return result.value;
+        }
+        return result;
+      });
+    }
+    return ValueNotifier(value);
+  }
```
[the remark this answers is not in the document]
andrewkolos: I think the risk only exists if the user's LLM is compromised, no? We could add an arbitrary recursion depth limit I suppose, but it's not critical.

packages/a2ui_core/lib/src/primitives/data_path.dart
```diff
@@ -0,0 +1,82 @@
...
+    if (normalized.isEmpty) {
+      return DataPath([]);
+    }
+
+    final List<String> segments = normalized.split('/').map((s) {
+      return s.replaceAll('~1', '/').replaceAll('~0', '~');
+    }).toList();
+
+    return DataPath(segments);
+  }
```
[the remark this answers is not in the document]
andrewkolos: > It treats / as the root path (empty segments), whereas in RFC 6901, / represents a path with one empty segment [""]. The root is an empty string "".

The A2UI protocol intentionally treats / as the root/whole document (see [a2ui_protocol.md:](https://github.com/google/A2UI/blob/059592efd73c032dd5bd7e7d6429efacccef8a76/specification/v0_9/docs/a2ui_protocol.md?plain=1#L249) "If omitted, or set to '/', refers to the entire data model."). This differs from RFC 6901 where / would mean the empty-string key, but it's the a2ui spec's deliberate choice... DataModel.set('/', value) and MessageProcessor both depend on this behavior.

> It strips trailing slashes, which prevents distinguishing between /foo and /foo/ (the latter should have a second empty segment).

The trailing slash stripping is an intentional defensive measure--LLM-generated JSON Pointers won't (intentionally) produce /foo/, and silently normalizing it is safer than creating a segment for a key named "".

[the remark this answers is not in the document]
andrewkolos: If I can remember, I'll add in code comments here so future readers/coding agents can better understand and judge the rationale used here.

.github/workflows/flutter_packages.yaml
```diff
@@ -172,4 +172,4 @@ jobs:
         working-directory: ${{ matrix.package.path }}
         run: |
           dart pub global activate layerlens
-          layerlens --fail-on-cycles  --except "lib/src/schema"
+          layerlens --fail-on-cycles  --except "lib/src/schema" --except "lib/src/core"
```
andrewkolos, line 175: The `core/` directory contains the A2UI "Core Data Layer" ([renderer guide](https://github.com/google/A2UI/blob/main/specification/v0_9/docs/renderer_guide.md?plain=1) Section 3): catalog API definitions, state models, and context classes. These have an inherent mutual dependency: `FunctionImplementation.execute()` receives a DataContext parameter, and `DataContext` invokes catalog functions to resolve `FunctionCall` expressions within `DynamicValue`s. This is mutually recursive by design--the same pattern exists in the web_core reference implementation (catalog/types.ts <-> rendering/data-context.ts). layerlens flags it.

Excluding `core/` from intra-directory cycle checking is appropriate because these types are a single cohesive unit per the spec. I don't believe this is flagging a legitimate concern one typically sees with a dependency cycle: there's no cycle in the construction graph, there's no infinite loop, there's no testability issue--the data context and function invoking behavior are both independently testable.

packages/a2ui_core/lib/src/common/cancellation.dart
```diff
@@ -0,0 +1,58 @@
...
+  /// Whether the operation has been cancelled.
+  bool get isCancelled => _isCancelled;
+
+  /// Cancels the operation.
+  void cancel() {
+    if (_isCancelled) return;
+    _isCancelled = true;
+    for (final void Function() listener in _listeners) {
+      listener();
+    }
```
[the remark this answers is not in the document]
andrewkolos: Fixed in [Fix CancellationSignal concurrent modification during cancel](https://github.com/flutter/genui/pull/857/commits/a1b16170dad862468c08a597e17e422e7bd21108)

packages/a2ui_core/lib/src/processing/basic_functions.dart
```diff
@@ -0,0 +1,53 @@
...
+
+    if (parts.isEmpty) return '';
+    if (parts.length == 1 && parts[0] is String) return parts[0];
+
+    return ComputedNotifier(() {
+      final Iterable<String> resolvedParts = parts.map((part) {
+        if (part is String) return part;
+        final ValueListenable<Object?> listenable = context.resolveListenable(
+          part,
+        );
```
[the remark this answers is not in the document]
andrewkolos: See https://github.com/flutter/genui/pull/857#discussion_r3042065935

packages/a2ui_core/lib/src/core/contexts.dart
```diff
@@ -0,0 +1,135 @@
...
+    }
+    if (value is Map && value.containsKey('call')) {
+      final call = FunctionCall.fromJson(Map<String, dynamic>.from(value));
+      return ComputedNotifier(() {
+        final args = <String, dynamic>{};
+        for (final MapEntry<String, dynamic> entry in call.args.entries) {
+          final ValueListenable<Object?> resolved = resolveListenable(
+            entry.value,
+          );
+          args[entry.key] = resolved.value;
```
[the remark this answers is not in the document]
andrewkolos: Worth re-consideration in a follow-up PR. I think the risk of this issue is low, and the `web_core` reference implementation shares the same issue. A proper fix might require rethinking how `resolveListenable` works, which would require a small but non-trivial change to the renderer spec that's beyond the scope of this PR.

packages/a2ui_core/lib/src/primitives/data_path.dart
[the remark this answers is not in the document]
andrewkolos: I just noticed that the A2UI spec got an update that partially addresses the bot's original concern: https://github.com/google/A2UI/blame/059592efd73c032dd5bd7e7d6429efacccef8a76/specification/v0_9/docs/a2ui_protocol.md#L395.

Still, ~~two~~ one of my implementation choices identified by the review bot are still not accounted for:
a. ~~`/` being treated as root instead of a key with the value of the empty string (`""`)~~
b. the trailing slash-stripping behavior (`/foo/` is reduced to `['foo']` instead of `[foo, ""]`)

I will try to submit an update to the spec to address these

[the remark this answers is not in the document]
andrewkolos: Upon further consideration, I think that the prior-mentioned spec update fully covers both of the bot's concerns. It covers concern 1 explicitly and concern 2 implicitly (when paired with the fact that empty string keys have no semantic meaning outside of arguably very specific academic use cases)

packages/a2ui_core/lib/src/primitives/reactivity.dart
```diff
@@ -0,0 +1,164 @@
...
+      return;
+    }
+    notifyListeners();
+  }
+}
+
+/// A derived notifier that automatically tracks and listens to other
+/// [ValueListenable] dependencies, recalculating its value only when they
+/// change.
+class ComputedNotifier<T> extends ValueNotifier<T> {
```
jacobsimionato: So these extensions basically allow creating signal or rx-like chains from ValueNotifier.

For this use case, I'm wondering if there might be other libraries already in use in the Dart ecosystem that could be a good fit. We used rxjs observables for a while on web, but it wasn't as good a fit as signals because it only represents value updates, not really the current value. I wonder if we should consider depending on https://pub.dev/packages/signals ?

packages/a2ui_core/lib/src/core/catalog.dart
```diff
@@ -0,0 +1,47 @@
...
+/// A definition of a UI component's API.
+abstract class ComponentApi {
+  String get name;
+  Schema get schema;
+}
+
+/// A definition of a UI function's API.
+abstract class FunctionApi {
+  String get name;
+  String get returnType;
```
jacobsimionato: Maybe return type should be an enum, like in https://github.com/flutter/genui/blob/29daadd5bace29ed934836d9443f0da656bc569e/packages/genui/lib/src/model/client_function.dart#L80 ?

packages/a2ui_core/lib/src/core/catalog.dart
jacobsimionato, line 1: Side comment: Have you tried getting an agent to compare this API surface to what's inside the existing Gen UI SDK, to try to match up the equivalent classes and see what APIs are changing? It'd be nice to get a sense of how awkward the transition will be. It could make sense to do some renaming / refactoring up front to the Gen UI SDK to reduce the migration diff, e.g. rename any parameters of the Catalog / CatalogItem / Function types etc, so that a2ui_core is more like a drop-in replacement.

packages/a2ui_core/lib/src/core/catalog.dart
```diff
@@ -0,0 +1,47 @@
...
+import 'contexts.dart';
+
+/// A definition of a UI component's API.
+abstract class ComponentApi {
+  String get name;
+  Schema get schema;
+}
+
+/// A definition of a UI function's API.
+abstract class FunctionApi {
```
jacobsimionato, line 37: I noticed that in Gen UI SDK, we currently have a description parameter on the function too. Is that part of the A2UI standard? I can't remember. We should try to align the A2UI specification and TS and Dart implementations to whatever makes the most sense. Perhaps the description should just be included on the argumentSchema? Or maybe a top-level description is not possible there, so it should be separate.

packages/a2ui_core/lib/src/core/surface_model.dart
```diff
@@ -0,0 +1,148 @@
...
+  void dispose() {
+    dataModel.dispose();
+    componentsModel.dispose();
+    _onAction.dispose();
+    _onError.dispose();
+  }
+}
+
+/// The root state model for the A2UI system.
+class SurfaceGroupModel<T extends ComponentApi> {
```
jacobsimionato: Let's put this in a separate file under `core/` or `model/` etc.

packages/a2ui_core/lib/src/core/surface_model.dart
```diff
@@ -0,0 +1,148 @@
...
+  final Map<String, SurfaceModel<T>> _surfaces = {};
+  final Map<String, void Function()> _actionForwarders = {};
+
+  final _onSurfaceCreated = ValueNotifier<SurfaceModel<T>?>(null);
```
jacobsimionato: This may have been from my prototype? But I don't think ValueNotifier is the right choice here, because that represents a specific value that can change, while conceptually, this should be notifying that different surfaces are being created or deleted. Maybe this should be a plain old callback? In web_core, we defined our own little observable utility for this - https://github.com/google/A2UI/blob/f9b732af5e0488dd1e2d12059e6dc1b9e69b077a/renderers/web_core/src/v0_9/common/events.ts#L30

Not sure if that makes sense here or if there is a good fit?

The same applies to some of the other models, e.g. ComponentModel etc.

packages/a2ui_core/lib/src/primitives/cancellation.dart
```diff
@@ -0,0 +1,59 @@
+// Copyright 2025 The Flutter Authors.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// A signal that can be used to cancel an operation.
+class CancellationSignal {
```
jacobsimionato, line 6: It feels like we're having to add a lot to ValueNotifier to get the overall signal-like behavior we need. Overall, we need some approach that offers:
- A) Representing the current value *and* updates to that value
- B) Synchronous processing rather than async (I think!)
- C) Ability to chain transformation functions together, which may depend on multiple inputs
- D) Ability for the ultimate consumer of the output to cancel their subscription or release the reference and have the chain of transformers, instance of function objects etc be destroyed.

Currently, we use ValueNotifier which offers A and B, then we add C and D. I wonder if there is a library out there that offers 3 or 4 of the properties out of the box?

packages/a2ui_core/lib/src/rendering/binder.dart
```diff
@@ -0,0 +1,421 @@
...
+
+class ChildNode {
+  final String id;
+  final String basePath;
+  ChildNode(this.id, this.basePath);
+
+  Map<String, dynamic> toJson() => {'id': id, 'basePath': basePath};
+}
+
+/// A framework-agnostic engine that transforms raw A2UI JSON payload
```
polina-c: It is not clear for me what is "payload configurations". 

Can you either define this term (in doc comments of some element or in go/genui-glossary ?) or use other words?

.github/workflows/flutter_packages.yaml
polina-c: Yes, there is no technical issues here. Just readability, and not for everyone.

Some engineers' dev velocity increases when there is no cycles, as the code is hierarchical.

And, yes, those engineers can easily clean up cycles here when they feel they are slowed down by the cycles presence.

packages/a2ui_core/lib/src/primitives/reactivity.dart
polina-c, line 1: Can you rename folder ‘common’ to ‘shared’ or ‘primitives’ as:
1. ‘shared’ or ‘primitives’ are used more in dash projects
2. ‘common’ is anti-pattern as it is ambiguous: it sometimes mean high level orchestrating helpers and sometimes low level primitives

packages/a2ui_core/lib/src/primitives/reactivity.dart
polina-c: As I know variations of such notifier are copied from package to package.

I expect dependency on signals to slow down dev velocity, because we will have to maintain copy of signals in G3.

.github/workflows/flutter_packages.yaml
polina-c: For example, this diagram becomes harder to understand, because of cycles:

Screenshot 2026-04-08 at 12 16 29 PM

polina-c:
Thank you for creating good test coverage!

packages/a2ui_core/lib/src/core/contexts.dart
```diff
@@ -0,0 +1,135 @@
...
+// found in the LICENSE file.
+
+import '../common/reactivity.dart';
+import 'catalog.dart';
+import 'common.dart';
+import 'component_model.dart';
+import 'data_model.dart';
+import 'surface_model.dart';
+
+/// A contextual view of the main DataModel.
```
polina-c: Can you give more details:
- what is contextual view?
- how it is used?

packages/a2ui_core/lib/src/core/contexts.dart
```diff
@@ -0,0 +1,135 @@
...
+      surface,
+      childModel,
+      basePath: basePath ?? dataContext.path,
+    );
+  }
+}
+
+extension CatalogInvokerExtension on Catalog {
+  /// Helper to invoke functions.
+  Object? invoker(String name, Map<String, dynamic> args, DataContext context) {
```
polina-c: Does 'invoker' return function that actually invokes function? Or result of invocation?

If result, should the method be named 'invoke'?

Can you add comment that clarifies this?

Maybe say 'client functions' instead of 'functions', to remove ambiguity?

packages/a2ui_core/lib/src/core/contexts.dart
```diff
@@ -0,0 +1,135 @@
...
+
+  ComponentContext(this.surface, this.componentModel, {String? basePath})
+    : dataContext = DataContext(surface, basePath ?? '/');
+
+  /// Dispatches an action from the component.
+  Future<void> dispatchAction(Map<String, dynamic> action) {
+    return surface.dispatchAction(action, componentModel.id);
+  }
+
+  /// Resolves a child component's context.
```
polina-c: `resolve` may mean many things. 

Maybe 'Returns ...'?

packages/a2ui_core/lib/src/core/contexts.dart
```diff
@@ -0,0 +1,135 @@
...
+/// Context provided to components during rendering.
+class ComponentContext {
+  final SurfaceModel surface;
+  final ComponentModel componentModel;
+  final DataContext dataContext;
+
+  ComponentContext(this.surface, this.componentModel, {String? basePath})
+    : dataContext = DataContext(surface, basePath ?? '/');
+
+  /// Dispatches an action from the component.
```
polina-c, line 122: It is not clear what meand 'dispatch' here?

Will the action be executed as result? Or just recieved by some other component?

Can you clarify?

packages/a2ui_core/lib/src/core/contexts.dart
```diff
@@ -0,0 +1,135 @@
...
+/// A contextual view of the main DataModel.
+class DataContext {
+  final SurfaceModel surface;
+  final String path;
+
+  DataContext(this.surface, this.path);
+
+  DataModel get dataModel => surface.dataModel;
+
+  /// Resolves a path against this context.
```
polina-c: Can you add more details what 'resolves' mean here?

Will it return absolute path?

May be examples will help.

packages/a2ui_core/lib/src/core/contexts.dart
```diff
@@ -0,0 +1,135 @@
...
+    if (relativePath.startsWith('/')) return relativePath;
+    if (relativePath == '' || relativePath == '.') return path;
+
+    final String base = path == '/'
+        ? ''
+        : (path.endsWith('/') ? path.substring(0, path.length - 1) : path);
+    return '$base/$relativePath';
+  }
+
+  /// Synchronously evaluates a dynamic value.
```
polina-c: Can you explain in more details contract of this method?

What is expected for the input?

What will be output for which input?

packages/a2ui_core/lib/src/core/component_model.dart
```diff
@@ -0,0 +1,90 @@
...
+  final Map<String, ComponentModel> _components = {};
+  final _onCreated = ValueNotifier<ComponentModel?>(null);
+  final _onDeleted = ValueNotifier<String?>(null);
+
+  /// Fires when a new component is added to the model.
+  ValueListenable<ComponentModel?> get onCreated => _onCreated;
+
+  /// Fires when a component is removed, providing the ID of the deleted
+  /// component.
+  ValueListenable<String?> get onDeleted => _onDeleted;
```
polina-c: onCreated and onDeleted:

ValueListenable is not right type here.

ValueListenable makes sense when there is notion of current value.

Should we use Stream<ComponentModel> and Stream<String> here?

packages/a2ui_core/lib/src/core/common.dart
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  Map<String, dynamic> toJson() => {
+    if (event != null) 'event': event,
+    if (functionCall != null) 'functionCall': functionCall!.toJson(),
+  };
+}
+
+/// A template for generating a dynamic list of children.
+class ChildListTemplate {
```
polina-c, line 75: I am missing explanation of what is "template". 

As there is no single class where this explanation can be added, should it be added to go/a2ui-glossary ? Or, if there will be many templates, we can create tarent class Template, and add explanation here.

However, I see that this class is not used. 

Can you explain target use case in doc comment?

packages/a2ui_core/lib/src/core/common.dart
```diff
@@ -0,0 +1,85 @@
...
+
+  Map<String, dynamic> toJson() => {
+    'call': call,
+    'args': args,
+    'returnType': returnType,
+  };
+}
+
+/// Triggers a server-side event or a local client-side function.
+class Action {
```
polina-c, line 49: Would it be better to create sealed class Action with two implementations: EventAction and FunctionAction?

polina-c (approved):
LGTM with comments

.github/workflows/flutter_packages.yaml
polina-c: So, balance your cost/value: if it takes more than ~15 min for you to make it cycle-free, yes, add exception for this folder.

packages/a2ui_core/lib/src/core/catalog.dart
andrewkolos: Oops, I forgot to come back to add these types proper. Thanks for the catch! Addressed in [Replace returnType string with A2uiReturnType enum](https://github.com/flutter/genui/pull/857/commits/7345376b61f582d745a3e82e3904518ce75836b6).

packages/a2ui_core/lib/src/core/surface_model.dart
andrewkolos: Done in [Split SurfaceGroupModel into its own file](https://github.com/flutter/genui/pull/857/commits/b66e0c54f3f14aec1bf6c682f89046648c1ded25)

packages/a2ui_core/lib/src/core/surface_model.dart
andrewkolos: Agreed. Addressed in [Add EventNotifier for discrete events](https://github.com/flutter/genui/pull/857/commits/553f468d35ef73e7bea6030d9373e0bd34a516ea). I followed your suggestion and added a `EventNotifier<T>`, modeled after web_core's `EventEmitter`. All event properties should be using it now.

packages/a2ui_core/lib/src/primitives/reactivity.dart
andrewkolos: I spent a few hours here, and I now believe it's worth taking on a dependency [on package:signals] here. I kept running into edge cases (or ballooning code complexity) trying to simulate signal model using push-based model like ValueNotifier.

> As I know variations of such notifier are copied from package to package.

I did consider a custom (reduced) implementation of package:signals, but even that would be decently complex (version tracking and managing topology of computed values). I also like the idea of using a heavily field-tested library that can also give us things like optimized/eager listener disposal, cycle detection, and graceful handling of edge cases such as a computed evaluation callback creating additional computeds.

I've added package:preact_signals (which package:signals wraps), which is just a Dart port of the same Preact signals used in the TypeScript renderer IIRC. We can upgrade to package:signals if needed when we proceed with migrating package:genui onto package:a2ui_core.

packages/a2ui_core/lib/src/primitives/cancellation.dart
andrewkolos: See https://github.com/flutter/genui/pull/857#discussion_r3060602117. I think we should just use package:preact_signals

packages/a2ui_core/lib/src/core/common.dart
jacobsimionato: I think this class directly mirrors the structure of the underlying JSON, and that is simpler to understand overall. Though I agree that purely from a Dart perspective, a sealed class is nicer.

packages/a2ui_core/lib/src/core/common.dart
jacobsimionato: Templates are an A2UI concept - see https://github.com/google/A2UI/blob/e30a9556cc3a138f6bafbaa6530627ebe8bb1e4a/specification/v0_9/docs/a2ui_protocol.md#collection-scopes-relative-paths

Perhaps they could be in the glossary, though I think the spec explains them well enough.

packages/a2ui_core/lib/src/core/contexts.dart
```diff
@@ -0,0 +1,150 @@
...
+    }
+    if (value is List) {
+      return value.map(resolveSync).toList();
+    }
+    return value;
+  }
+
+  /// Returns a reactive signal that re-evaluates a dynamic value
+  /// whenever its underlying data dependencies change.
+  ReadonlySignal<Object?> resolveListenable(Object? value) {
```
jacobsimionato, line 76: resolveSignal ?

jacobsimionato: Or `listen`?

packages/a2ui_core/lib/src/rendering/binder.dart
```diff
@@ -0,0 +1,402 @@
...
+
+class BehaviorNode {
+  final Behavior type;
+  final Map<String, BehaviorNode>? shape;
+  final BehaviorNode? element;
+
+  BehaviorNode(this.type, {this.shape, this.element});
+}
+
+class ChildNode {
```
jacobsimionato, line 24: Maybe `ChildNodeRef` might be a better name? In future we might introduce the concept of actual nodes which contain the resolved props etc for the node.

jacobsimionato (approved):

polina-c:
@andrewkolos 

I see you switched from using flutter-like notifiers to signals.

It is a significant deviation from original plan, so we want to make this decision at team level.

In general, if you make a big change after getting LGTM, it is good to re-ask for review (or, when change is urgent, to merge PR and ask for post-review and be ready to address comments in the following PRs).

In [this comment](https://github.com/flutter/genui/pull/857#discussion_r3060602117) I do not see reasons not to use flutter-like notifiers.

May I ask you to outline pros and cons for flutter-like notifiers vs signals and discuss them with the team?

andrewkolos:
> @andrewkolos
> 
> I see you switched from using flutter-like notifiers to signals.
> 
> It is a significant deviation from original plan, so we want to make this decision at team level.
>
> In general, if you make a big change after getting LGTM, it is good to re-ask for review (or, when change is urgent, to merge PR and ask for post-review and be ready to address comments in the following PRs).
> 
> In [this comment](https://github.com/flutter/genui/pull/857#discussion_r3060602117) I do not see reasons not to use flutter-like notifiers.
> 
> May I ask you to outline pros and cons for flutter-like notifiers vs signals and discuss them with the team?

Understood. I'll try to be more liberal in re-requesting reviews after pushing changes. Leaving a summary of a decision and treating the lack of any quick reply as agreeance was a bit hasty on my part. 

I'm currently wrapping up some work on google/a2ui from last week, but I'll circle back to this right after to reconstruct my full thought process and share it.

PR #889, opened 2026-04-24
chore: bump versions for release

jacobsimionato:
### Description of Changes

This PR prepares the workspace for a new release cycle by bumping package versions and synchronizing their cross-dependencies. 

Specifically:
- Bumped `genui` to `0.9.0`.
- Bumped `genui_a2a` to `0.9.0`.
- Initialized and bumped the new `a2ui_core` package to `0.1.0`.
- Updated all intra-repo dependencies across the workspace (including all example apps) to point to these new versions.
- Updated `CHANGELOG.md` files to document the features, fixes, and breaking changes included in these bumps.
- Fixed an issue in some example app CMake and Swift files.

### Rationale

These bumps reflect the latest batch of updates merged into `main`, including the structural reorganization that extracted protocol primitives into `a2ui_core`, which required a breaking minor version bump for the dependent packages before publishing.

### Testing/Running Instructions

1. Checkout the branch locally.
2. Run `dart pub get` and `flutter pub get` to ensure all dependencies resolve correctly.
3. Verify that the example apps build successfully (e.g. `cd examples/travel_app && flutter build linux`).

PR #890, opened 2026-04-24
chore: bump versions for release (2)

polina-c:
Successor of https://github.com/flutter/genui/pull/889, with version of a2ui_core changed to 0.0.1-dev002, that is one way to address [the comment](https://github.com/flutter/genui/pull/889#discussion_r3140460885).


jacobsimionato (approved):
LGTM - thanks for coming up with a better approach here.

If I could turn back time, I think it might have been ideal to avoid adding a dependency from `genui` to `a2ui_core` before we were ready to release `a2ui_core`, seeing as now if we want to release `genui` we *need* to release `a2ui_core`. But I think we need to cross this bridge eventually, so I like your solution here of releasing `a2ui_core`, but with a version label that makes it clear that it's not ready for widespread use yet.

polina-c:
> LGTM - thanks for coming up with a better approach here.
> 
> If I could turn back time, I think it might have been ideal to avoid adding a dependency from `genui` to `a2ui_core` before we were ready to release `a2ui_core`, seeing as now if we want to release `genui` we _need_ to release `a2ui_core`. But I think we need to cross this bridge eventually, so I like your solution here of releasing `a2ui_core`, but with a version label that makes it clear that it's not ready for widespread use yet.

I agree. Noted for future. It was me - I could not put Flutter adapter to the core library, because it is pure Dart. So I've put it to Flutter library `genui`, not thinking that it will block release.

PR #894, opened 2026-04-30
Delete example `verdure` and platforms that we do not test for.

polina-c:


jacobsimionato:
Hey can you chat to Alan about Verdure? I think I agree with you we can just delete it now that it's served its purpose, but he might have good reasons to keep it. Some other options would be:
- Test it first in case it works, in which case we can move it to a separate repo and pin the SDK versions that it uses
- Get it working again

jacobsimionato:
Re the other apps:
Can we delete windows and linux but keep iOS and Android and try to support them? They are the major platforms that people associate with Flutter and iOS was working last time I checked, so I think we should continue to support them.

I think macos/web/ios/android is a good set to aim for (macos purely for our own dev purposes). WDYT?

polina-c:
Let's have discussion on the team.

It is cost to support many platforms.

Not sure we want to pay this cost.

gspencergoog:
Since Verdure is fixed now, can we close this PR?

PR #895, opened 2026-05-04
Cleanup listenables as we decided to go with signals.

polina-c:


andrewkolos (approved):
Thanks for cleaning this up 👍

PR #905, opened 2026-05-10
Improve simple chat and make some cross-repo cleanups.

### LINKED ISSUE #808: Code of simple_chat is over-complicated.
Code of [simple_chat](https://github.com/flutter/genui/tree/main/examples/simple_chat/lib) should be compact, easy to read and deal with a small number of clearly defined concepts of GenUI SDK.

In addition, we need to setup some continuous validation for this simplicity, because we [already worked](https://github.com/flutter/genui/issues/114) to simplify simple_chat before.

polina-c:
Contributes to https://github.com/flutter/genui/issues/907
Fixes https://github.com/flutter/genui/issues/808

## Simple chat

1.  Factor out agent from transport to improve readability
2. Rename code elements to match terminology
3. Inspired by @yjbanov 's recent demo, introduced modes: text-only, basic catalog, custom catalog (default)
4. Stop using class Conversation
5. Hard code default initial user message to enable faster demos

## Cross-repo cleanups

1. Fix integration tests and add them to CI
2. Rename 'eval' to 'e2e'
3. Rename getApiKey to apiKey, because 'get' prefix in methods is dart antipattern
4. Add BasicCatalogItems.asNoAssetCatalog, that excludes items that require assets (audio, video, images)
 
## Video

https://github.com/user-attachments/assets/469fb2cf-09cf-463c-8c9c-b9c0cb39203b


.github/workflows/flutter_packages.yaml
```diff
@@ -50,7 +50,8 @@ jobs:
           GH_TOKEN: ${{ github.token }}
         run: |
           # Find all directories containing pubspec.yaml in packages, tools, and examples.
-          ALL_DIRS=$(find packages examples tool -name pubspec.yaml -not -path "*/.dart_tool/*" -not -path "examples/eval/*" -exec dirname {} \;)
+          # examples/e2e is excluded, because it requires GEMINI_API_KEY and is exercised in other workflow
```
ditman, line 53: in other workflow -> in the e2e.yaml workflow ?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -11,6 +11,7 @@ _List which issues are fixed by this PR. For larger changes, raising an issue fi
 - [ ] I read the [Contributors Guide].
 - [ ] I have added sample code updates to the [changelog].
 - [ ] I updated/added relevant documentation (doc comments with `///`).
+- [ ] If my PR is a fork PR, I've checked that [e2e tests] passed.
```
ditman, line 14: Can tests be run by end users? We have this same message in google/A2UI, and it's not super relevant unless flutter code is being touched (?)

.vscode/launch.json
```diff
@@ -28,6 +28,15 @@
         "--dart-define=GEMINI_API_KEY=${env:GEMINI_API_KEY}"
       ]
     },
+    {
+      "name": "verdure_client",
+      "cwd": "examples/verdure/client",
+      "request": "launch",
+      "type": "dart",
+      "args": [
+        "--dart-define=GEMINI_API_KEY=${env:GEMINI_API_KEY}"
```
ditman, line 37: Is this --dart-define needed? I'd expect if there's a GEMINI_API_KEY already set in the environment that it is available in the flutter program without doing a --dart-define again?

examples/composer/lib/api_key.dart
```diff
@@ -8,20 +8,20 @@ const String _geminiApiKey = String.fromEnvironment('GEMINI_API_KEY');
 
 String? debugApiKey;
 
-String getApiKey() {
+String apiKey() {
```
ditman, line 11: Why this renaming? Shouldn't a function name contain a verb?

examples/e2e/test/simple_chat_test.dart
```diff
@@ -0,0 +1,187 @@
...
+    );
+
+    tester.verifyEvents();
+
+    tester.failIfIssuesFound();
+  }, timeout: const Timeout(Duration(minutes: 4)));
+}
+
+/// Helper class to manage a chat session from simple chat example.
+class _ChatSessionTester {
```
ditman: Consider splitting this class to a separate lib so the test file remains small.

examples/simple_chat/assets/climbing/296x3072x2048.jpg
ditman, line 1: Where are all these images coming from?

examples/e2e/test/simple_chat_test.dart
```diff
@@ -0,0 +1,187 @@
...
+  'Add a task "Hire architect" to my todo list',
+  'Mark the task "Hire architect" as completed',
+  'Remove the task "Hire architect"',
+  'Clear my todo list',
+];
+
+void main() {
+  test('Model respects configuration of prompt builder '
+      'in the simple chat example.', () async {
+    final tester = _ChatSessionTester();
```
ditman: I know this file is probably moved from the earlier location and unchanged in this PR, but consider renaming `tester` to something else.

In flutter integration tests (`testWidgets`...) `tester` is a very specific API that is everywhere.

I'd call this `chatSessionTester` or something like that (see `app_test.dart` below)

examples/simple_chat/integration_test/app_test.dart
```diff
@@ -134,9 +121,14 @@ Future<void> _runTestForSample(
   await tester.tap(find.byIcon(Icons.send));
   await tester.pump(); // Start processing
 
-  // Wait for response and rendering
+  // Wait for response and rendering.
   // The FakeAiClient splits it into chunks with delays.
-  await tester.pumpAndSettle();
+  // We can't use pumpAndSettle() because some catalog widgets (e.g. Image's
+  // loadingBuilder shows a CircularProgressIndicator) have indeterminate
+  // animations that never settle.
```
ditman: Do they never settle because the network is not available, or because the image that we're attempting to load has a bad URL?

examples/simple_chat/lib/chat_session.dart
```diff
@@ -23,10 +28,24 @@ If there is no way to itemize all the options, either use the component '${Basic
...
+final Catalog _customCatalog = _basicCatalog.copyWith(
+  systemPromptFragments: [
+    '''
+If the user is asking about climbing locations, use the 'listClimbingLocations' tool to get a list of climbing locations.
+Always use the component named '${climbingLocationItem.name}' to display the locations. The '${climbingLocationItem.name}' component already includes a 'Learn more' button; do not add any extra submit/confirmation buttons next to it.
+When the user clicks 'Learn more' on a '${climbingLocationItem.name}', a UI action named 'learnMoreAboutLocation' will be sent with the location's identifier and name in its context. Respond with detailed information about that specific location.
+
+When user asks about climbing locations, never use other components.
+''',
+    ..._basicCatalog.systemPromptFragments,
```
ditman, line 54: should the default systemprompts be placed before the custom catalog fragments?

examples/simple_chat/lib/chat_session.dart
```diff
@@ -23,10 +28,24 @@ If there is no way to itemize all the options, either use the component '${Basic
   ],
 );
 
-final PromptBuilder _promptBuilder = PromptBuilder.chat(
-  catalog: _catalog,
+final Catalog _customCatalog = _basicCatalog.copyWith(
+  systemPromptFragments: [
+    '''
```
ditman: It'd be nice that all the system prompts in this file were extracted to `const SOMETHING_PROMPT = '''` at the top of the file, so they're close together, and then at the bottom the code is just "glue" to wire the system prompts to their correct locations below.

examples/simple_chat/README.md
```diff
@@ -33,3 +33,11 @@ The application's logic is contained almost entirely within `lib/chat_session.da
 ## Getting Started
 
 Follow the instructions in [Running the app with a Gemini key](../../docs/usage/run_app_with_gemini_key.md).
+
+## Video
+
+https://github.com/user-attachments/assets/469fb2cf-09cf-463c-8c9c-b9c0cb39203b
```
ditman, line 39: This demo looks super good. Kudos!

packages/genui/lib/src/catalog/basic_catalog.dart
```diff
@@ -105,7 +105,16 @@ abstract final class BasicCatalogItems {
...
+  /// Creates a basic catalog without items that require assets.
+  ///
+  /// This is useful for the app, that do not work with images, audio or video.
+  static Catalog asNoAssetCatalog({
+    List<String> systemPromptFragments = const [],
+  }) => asCatalog(
+    systemPromptFragments: systemPromptFragments,
+  ).copyWithout(itemsToRemove: [audioPlayer, image, video]);
+
+  /// Creates a catalog containing all basic catalog items.
```
ditman: Not a fan of exposing an uglier version of the catalog. I think this should be added by the application itself, rather than our API.

ditman (approved):
Some small comments. Nothing blocking. My biggest concern is exposing the "no asset" version of the basic catalog through its API (it's hard to remove later, and it's something that apps can do very easily themselves when they need it)

.github/PULL_REQUEST_TEMPLATE.md
polina-c: In a2ui it is relevant if python code is touched, because service is in python. 
We have all kinds of dependencies, and we may add more e2e tests. Added condition for changed code. Thank you!

examples/composer/lib/api_key.dart
polina-c: See https://dart.dev/effective-dart/design#avoid-starting-a-method-name-with-get

examples/simple_chat/assets/climbing/296x3072x2048.jpg
polina-c: yegor generated them for cloud next

examples/e2e/test/simple_chat_test.dart
polina-c: splitted

examples/e2e/test/simple_chat_test.dart
polina-c: renamed to sessionTester

examples/simple_chat/integration_test/app_test.dart
polina-c: CircularProgressIndicator messes up somehow with pumpAndSettle. This fix is suggested by agent. Updated comment. I do not know details and i do not feel it is worth to spend time investigation.

examples/simple_chat/lib/chat_session.dart
polina-c: Not sure. The example works this way well, and it may break if i change the prompt. Leaving it as is.

packages/genui/lib/src/catalog/basic_catalog.dart
polina-c: Asset related catalog items require addisional configuration of assets. They cannot be just plugged in. So, it results in extra troubleshooting for early adopters. 

Added more comments to both methods to make it clear. Thank you!

PR #912, opened 2026-05-12
fix(a2ui_core/data_model): notify descendant watches on root set

andrewkolos:
Part of work on https://github.com/flutter/genui/issues/811

## Bug
In a2ui_core's `DataModel._notifyPathAndRelated`, `set('/', value)` would not notify subscribers at descendant paths (`/foo`, `/foo/bar`, etc.). Root path was treated as `'/'` in a `startsWith('$normalizedPath/')` check, which became `startsWith('//')` and never matched. Discovered while working on a2ui_core migration work.

## Fix
Replaces the ad-hoc string-prefix arithmetic with a small `_descendantPrefix` helper that isolates the root special case. The body of `_notifyPathAndRelated` can then check that one of three simple conditions is satisfied: equal, entry is descendant of changed, or changed is descendant of entry. Adds some test cases, which were verified as failing prior to the fix.

Unrelated cleanup: also tightens `Map`/`List` type checks in path traversal to `Map<Object?, Object?>` / `List<Object?>` so internal code propagates `Object?` instead of `dynamic`.

## Pre-launch Checklist
- [x] I signed the [CLA].

[changelog]: ../CHANGELOG.md

### DIFF
--- a/packages/a2ui_core/lib/src/core/data_model.dart
+++ b/packages/a2ui_core/lib/src/core/data_model.dart
@@ -149,6 +149,13 @@ class DataModel {
   void _notifyPathAndRelated(DataPath dataPath) {
     final normalizedPath = dataPath.toString();
 
+    if (dataPath.isEmpty) {
+      for (final String entryPath in _signals.keys.toList()) {
+        _getAndNotify(entryPath);
+      }
+      return;
+    }
+
     // Notify all active signals that are related to this path
     for (final String entryPath in _signals.keys.toList()) {
       if (entryPath == '/' || entryPath == '') {
--- a/packages/a2ui_core/lib/src/core/data_model.dart
+++ b/packages/a2ui_core/lib/src/core/data_model.dart
@@ -16,7 +16,7 @@ const int maxAutoVivifyIndex = 10000;
 /// It handles JSON Pointer path resolution and reactive signal management.
 class DataModel {
   Object? _data;
-  final Map<DataPath, WeakReference<Signal<Object?>>> _signals = {};
+  final Map<String, WeakReference<Signal<Object?>>> _signals = {};
 
   DataModel([Object? initialData]) : _data = initialData ?? <String, Object?>{};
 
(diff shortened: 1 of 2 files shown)


packages/a2ui_core/lib/src/core/data_model.dart
```diff
@@ -147,25 +146,19 @@ class DataModel {
...
-        _getAndNotify(entryPath);
-      } else if (normalizedPath.startsWith('$entryPath/')) {
-        _getAndNotify(entryPath);
-      } else if (entryPath.startsWith('$normalizedPath/')) {
+      if (changedPath == entryPath ||
+          entryPath.startsWith(changedDescendantPrefix) ||
+          changedPath.startsWith(_descendantPrefix(entryPath))) {
         _getAndNotify(entryPath);
       }
     }
```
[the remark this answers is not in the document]
andrewkolos: Bot is stuck hyper-fixated in an over-optimization loop.

1. I assume the Dart compiler is going to do the inline optimization automatically.
2. The string interpolation for the root path case is already skipped?
3. _N_ is signals per data model, this is not a large number in most/all applications.

packages/a2ui_core/lib/src/core/data_model.dart
```diff
@@ -28,9 +28,9 @@ class DataModel {
     Object? current = _data;
     for (final String segment in dataPath.segments) {
       if (current == null) return null;
-      if (current is Map) {
+      if (current is Map<Object?, Object?>) {
```
polina-c: Why not Map<String, Object?> in this case?

polina-c (approved):
LGTM with one question

packages/a2ui_core/lib/src/core/data_model.dart
polina-c: It would make sense to comment what is 'current'.

andrewkolos: Both were artefacts copied from the genui package's implementation. I've addressed them. Thanks for the catch!

PR #915, opened 2026-05-13
fix(schema): add missing pluralize and openUrl to clientFunctions

gspencergoog:
# Description

These functions were defined as standard functions in the A2UI v0.9 specification and implemented in basic_functions.dart, but were missing from the A2uiSchemas.clientFunctions() schema definition. Adding them improves compliance with the spec.

Also fixes https://github.com/flutter/genui/issues/902 and fixes https://github.com/flutter/genui/issues/903 by adding missing fields to pluralize.


ditman (approved):
Where would the implementation for these functions go? Somewhere in the basic catalog?

gspencergoog:
> Where would the implementation for these functions go? Somewhere in the basic catalog?

They were already implemented in `basic_functions.dart`.  There were some missing fields there, though, so I added them.

PR #917, opened 2026-05-13
Release automation.

polina-c:
Contributes to https://github.com/google/A2UI/issues/1383.

Prerequisite: https://github.com/flutter/genui/pull/926

1. Add workflows to automate release process (off for now)
1. Document release process
2. Cleanup some documentation
3. Fix pub.dev warnings by removing checked-in git ignores
4. Fix release validation errors 
5. Document update story (with dependabot)
6. Remove release tool, because it is not needed any more
7. Update .github/dependabot.yaml to check packages that are not in root '/' -> '/**'
8. Pin analyser and comment the problem
9. Add CI validation for locally broken version constraints


docs/contributing/publishing.md
```diff
@@ -0,0 +1,52 @@
...
+[firehose rules](https://github.com/dart-lang/ecosystem/tree/main/pkgs/firehose).
+
+There are two CI workflows that enable this automation:
+
+1. [post_summaries.yaml](../../.github/workflows/post_summaries.yaml)
+2. [publish.yaml](../../.github/workflows/publish.yaml)
+
+## Making PR passing `publish / validate`
+
+In general, the job [publish / validate](https://github.com/flutter/genui/actions/runs/25936562918) checks if all pub.dev packages are ready for publishing.
```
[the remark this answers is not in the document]
polina-c: nope

docs/contributing/publishing.md
```diff
@@ -0,0 +1,52 @@
...
+## Package categories
+
+`pub.dev` packages in this repo fall into three categories:
+
+1. **Not intended to be published**: they have `publish_to: none` in their `pubspec.yaml`.
+2. **Mono-repo packages**: they have `resolution: workspace` in their `pubspec.yaml`, and are released together, in lock-step, with the same version number.
+3. **Independent packages**: they don't have `publish_to` and `resolution` fields in their `pubspec.yaml`. They are released independently.
+4. **Not yet published packages**: they have `resolution: workspace` and `release: none` in their `pubspec.yaml`.
+
+## `-dev` vs non-`-dev` (production ready) versions
```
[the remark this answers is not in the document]
polina-c: nope

docs/contributing/publishing.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Making PR passing `publish / validate`
+
+In general, the job [publish / validate](https://github.com/flutter/genui/actions/runs/25936562918) checks if all pub.dev packages are ready for publishing. 
+
+To make sure your PR passes this validation, follow [firehose rules](https://github.com/dart-lang/ecosystem/tree/main/pkgs/firehose).
+
+## Package categories
+
+`pub.dev` packages in this repo fall into three categories:
```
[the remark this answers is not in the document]
polina-c: removed number

docs/contributing/publishing.md
```diff
@@ -0,0 +1,55 @@
...
+
+1. **Not published**: `pubspec.yaml` contains `publish_to: none`. Workspace tools and example apps that are never pushed to pub.dev.
+2. **Not yet published**: the package's `version:` ends with a `-dev<N>` suffix (see "`-dev` vs non-`-dev`" below). Published to pub.dev only to reserve the name; not ready for general use yet.
+3. **Published**: any other package. Each has its own version cadence on pub.dev.
+
+Note: `resolution: workspace` in a `pubspec.yaml` is a tooling concern — it tells Dart to share dependency resolution and a lockfile with the monorepo, and it does **not** by itself imply anything about release cadence. A package can opt out of the workspace (omit `resolution: workspace`) to avoid circular dependencies or unrelated update churn while still being a published package.
+
+## `-dev` vs non-`-dev` (production ready) versions
+
+The packages code should be always release ready. That means:
```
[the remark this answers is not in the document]
polina-c: 'packages' is right

docs/contributing/publishing.md
```diff
@@ -0,0 +1,55 @@
...
+
+1. Use `-dev` version if **at least one** of the following statements is true:
+
+   1.1. The package is planned to be released in the future. In this case it is published with `-dev` suffix in order to reserve the package name.
+
+   1.2. The package's changes touch only non-publishable code or docs (like tests, tools, or not-publishable docs).
+
+   You can publish `-dev<number>` versions, if you need it for development.
```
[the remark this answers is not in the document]
polina-c: it looks ok as is

docs/contributing/publishing.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Package categories
+
+Packages in this repo fall into the following categories:
+
+1. **Not published**: `pubspec.yaml` contains `publish_to: none`. Workspace tools and example apps that are never pushed to pub.dev.
+2. **Not yet published**: the package's `version:` ends with a `-dev<N>` suffix (see "`-dev` vs non-`-dev`" below). Published to pub.dev only to reserve the name; not ready for general use yet.
+3. **Published**: any other package. Each has its own version cadence on pub.dev.
+
+Note: `resolution: workspace` in a `pubspec.yaml` is a tooling concern — it tells Dart to share dependency resolution and a lockfile with the monorepo, and it does **not** by itself imply anything about release cadence. A package can opt out of the workspace (omit `resolution: workspace`) to avoid circular dependencies or unrelated update churn while still being a published package.
```
gspencergoog: If a package opts out of the monorepo, however, it is then implicitly switching to using published packages only, and local changes won't be testable without manual editing of the pubspec files to use local paths, or publishing the dependent packages.

How does using the monorepo increase circular dependency problems?  It seems like it would make them better, not worse, since you're using the local versions of the code.  But maybe I don't understand the use case you're describing.

And what kind of "update churn"?   Since it doesn't imply anything about release cadence, I wouldn't think that would increase the churn.  Are you referring to churn in publishing other packages, or churn the package that is part of the monorepo?

docs/contributing/publishing.md
```diff
@@ -0,0 +1,74 @@
...
+
+The packages code should be always release ready. That means:
+
+1. Use `-dev` version if **at least one** of the following statements is true:
+
+   1.1. The package is planned to be released in the future. In this case it is published with `-dev` suffix in order to reserve the package name.
+
+   1.2. The package's changes touch only non-publishable code or docs (like tests, tools, or not-publishable docs).
+
+   You can publish `-dev<number>` versions, if you need it for development.
```
gspencergoog: What is the format of `<number>`?  Is it another semver number, or just an ordinal?  Is it zero padded, or does it have a dash?

e.g. which of these is it?:

- `0.1.0-dev1`
- `0.1.0-dev001`
- `0.1.0-dev-1`
- `0.1.0-dev-0.1.0'

docs/contributing/publishing.md
```diff
@@ -0,0 +1,74 @@
...
+
+1. Use `-dev` version if **at least one** of the following statements is true:
+
+   1.1. The package is planned to be released in the future. In this case it is published with `-dev` suffix in order to reserve the package name.
+
+   1.2. The package's changes touch only non-publishable code or docs (like tests, tools, or not-publishable docs).
+
+   You can publish `-dev<number>` versions, if you need it for development. 
+
+2. If your feature is partially implemented, hide the feature's code behind a false-by-default flag, and use **release-ready** version.
```
gspencergoog: Where are the flags specified?  Are they `--define` flags for the Dart compiler?  Or are they command line flags? Something else?

docs/contributing/publishing.md
```diff
@@ -0,0 +1,74 @@
...
+
+TODO(polina-c): add information, https://github.com/google/A2UI/issues/1383
+
+## How upgrade of dependencies (for both siblings and third party) happens?
+
+### For local development runs
+
+For packages with `resolution: workspace` in their pubspec.yaml, pub resolves every sibling from its local source directory — not from pub.dev, as long as its `version:` satisfies the consumer's constraint.
+
+If a local bump escapes that constraint (e.g. `^0.9.0` → `0.10.0`), update the consumer's `pubspec.yaml` in the same PR. Otherwise pub silently falls back to the published version on pub.dev.
```
gspencergoog: Can we have a check for this?

docs/contributing/publishing.md
```diff
@@ -0,0 +1,74 @@
...
+
+### For runs by external packages
+
+After a new version of a dependency (including sibling package in this repo) is published, this is how upgrade will happen:
+
+1. [Dependabot] detects the new version on pub.dev and opens a PR per dependency, bumping the constraint in each consuming `pubspec.yaml`. See [About Dependabot version updates] for details.
```
gspencergoog, line 77: Is there any way to add a new section to each `CHANGELOG.md` too?

docs/contributing/README.md
```diff
@@ -23,15 +27,21 @@ I have read and understood ./docs/contributing/README.md.
 1. Documentation in the repository (all .md files) should be clear, consistent, concise and up-to-date.
 2. Documentation should not contain details that are easy to infer from the code.
 3. If code does not match the documentation, there should be TODO comments in the code to signal the discrepancy should be resolved.
+4. For documentation use [sentence case for headings](https://developers.google.com/style/capitalization#capitalization-in-titles-and-headings).
 
-## Code reviews
+## Shell scripts
 
-Do not review pull requests when they are in draft state, unless explicitly requested by the author.
+To run a script in `tool/`, open the script in VSCode and press ⇧⌘B.
```
gspencergoog: Also add instructions for where to run the script from on the command line: not everyone uses VSCode, and that key binding is Mac-only, so you might include the default Linux keybinding too (which is likely to be the same as Windows).  They also might have changed the keybinding...

docs/contributing/README.md
```diff
@@ -23,15 +27,21 @@ I have read and understood ./docs/contributing/README.md.
...
 
-  ```bash
-  ./tool/run_all_tests_and_fixes.sh
-  ```
+## pubspec.lock files
+
+`pubspec.lock` files are not git ignored to make the bots faster.
+
+If you include `pubspec.lock` file to your PR, make sure to run `flutter pub upgrade`,
+when your Flutter is latest at beta channel.
```
gspencergoog:
```suggestion
when your Flutter is using the latest version of the beta channel (run `flutter channel beta && flutter upgrade` to make sure you're on the right one).
```

Because this is using the beta channel, we will occasionally be breaking the people developing using the stable channel.  (Whenever the beta channel is depending on some API that will break the stable channel) It won't be often, but it can happen.

We do have CI to prevent the packages themselves from breaking things this way, but a developer using stable might be broken.

Maybe we should also have a CI test that makes sure that any `pubspec.lock` changes match the beta channel? Shouldn't be too hard to enforce.

packages/genai_primitives/CHANGELOG.md
```diff
@@ -1,9 +1,5 @@
 # `genai_primitives` Changelog
 
-## 0.2.4 (in progress)
-
-- **Refactor**: Update core framework to v0.9 (#546dab9be).
```
gspencergoog, line 5: This comment is now lost.  When will it be added back in?  Shouldn't there be a section for in progress (-dev) changes?

tool/fix_copyright/pubspec.yaml
```diff
@@ -4,6 +4,7 @@
 
 name: fix_copyright
 description: A command line app to fix copyright headers.
+publish_to: none
 version: 0.1.0
```
gspencergoog: If this is publish: none, should it also have a -dev version number?

packages/json_schema_builder/.pubignore
```diff
@@ -1 +0,0 @@
-submodules/JSON-Schema-Test-Suite/**
```
gspencergoog, line 1: If you remove this, then the test suite will be re-published with the package, and that's not our code. It will also make the package a lot bigger, and it's not data that we need to publish.

Don't we need to keep this?

docs/contributing/publishing.md
polina-c: Thank you!

Yes this documentation was confusing and some sentences are wrong. Updated.

We still use advantages of monorepo. Added section to explain how.

docs/contributing/publishing.md
polina-c: Thank you. Updated doc.

docs/contributing/publishing.md
polina-c: We do not have it organized yet. Documented that it is not defined yet.

docs/contributing/publishing.md
polina-c: Good idea!

added

tool/fix_copyright/pubspec.yaml
polina-c: it should not have version number at all

deleted here and in other places

docs/contributing/publishing.md
polina-c: Not natively — Dependabot only edits the manifest. I filed a feature request upstream: https://github.com/dependabot/dependabot-core/issues/15057. For now the doc calls out that the `CHANGELOG.md` entry has to be added by the maintainer before merging.

polina-c: Filed an issue and added comment: https://github.com/dependabot/dependabot-core/issues/15057

packages/genai_primitives/CHANGELOG.md
polina-c: - Changed the comment to make sense in the context of the package genai_primitives, that is not concerned about genui.

- Bumped version

packages/json_schema_builder/.pubignore
polina-c: reverted

thank you!

docs/contributing/publishing.md
```diff
@@ -0,0 +1,85 @@
...
+
+Packages in this repo fall into the following categories:
+
+1. **Not published**: `pubspec.yaml` contains `publish_to: none`. Workspace tools and example apps that are never pushed to pub.dev.
+2. **Not yet published**: the package's `version:` ends with a `-dev<N>` suffix (see "`-dev` vs non-`-dev`" below). Published to pub.dev to reserve the name or to test the package; not ready for general use yet. Format: `0.1.0-dev002`.
+3. **Published**: any other package. Each has its own version cadence on pub.dev.
+
+## About `resolution: workspace`
+
+`resolution: workspace` in a `pubspec.yaml` is a tooling concern and it does **not** by itself imply anything about release cadence:
```
gspencergoog: Why is it a "concern"?  That implies that we don't really want to use it.  Or do you mean that it only is something for the tooling to worry about?

docs/contributing/publishing.md
gspencergoog: I don't see a description of the format of `<number>` in the `-dev<number>` suffix.

packages/genai_primitives/CHANGELOG.md
gspencergoog: But what about in-progress changes: Say I've committed something that isn't meant to be published yet, but I want to accumulate changelog entries for it when it *is* published.  How am I supposed to do that?  I would think there would be a `## 0.2.4-dev` entry here with those things, that gets converted to a `## 0.2.4` heading before it is published.

docs/contributing/publishing.md
polina-c: good catch

resolved

packages/genai_primitives/CHANGELOG.md
gspencergoog: But what about in-progress changes: Say I've committed something that isn't meant to be published yet, but I want to accumulate CHANGELOG entries for it when it *is* published.  How am I supposed to do that?  I would think there would be a `## 0.2.4-dev` entry here with those things, that gets converted to a `## 0.2.4` heading before it is published.

polina-c:
> I don't see a description of the format of <number> in the -dev<number> suffix.

Yes, user expects to see it here. Moved here from above.

docs/contributing/README.md
```diff
@@ -23,15 +27,22 @@ I have read and understood ./docs/contributing/README.md.
...
 
-## Code reviews
+## Shell scripts
 
-Do not review pull requests when they are in draft state, unless explicitly requested by the author.
+To run a script in `tool/`:
 
-## Key commands
+- If you are on mac and use VS Code, open the script and press `⇧⌘B` (see [.vscode/tasks.json](../../.vscode/tasks.json)).
+- Otherwise, run `sh <path/to/script.sh>` from any directory.
```
gspencergoog: This will work, but aren't they all set executable already, so you just need `<path/to/script.sh>`?  You could just say "Otherwise, you can invoke the scripts on the command line from any directory." since all I was trying to indicate was the sense that you don't *have* to invoke them from VSCode, and I assume most developers know how to run a shell script.

docs/contributing/publishing.md
```diff
@@ -0,0 +1,85 @@
...
+
+The packages code should be always release ready. That means:
+
+1. Use `-dev` version (format `0.1.0-dev002`) if **at least one** of the following statements is true:
+
+   1.1. The package is planned to be released in the future. In this case it is published with `-dev` suffix in order to reserve the package name.
+
+   1.2. The package's changes touch only non-publishable code or docs (like tests, tools, or not-publishable docs).
+
+   You can publish `-dev<number>` versions, if you need it for development.
```
gspencergoog: Maybe:
```suggestion
   You can publish `-dev<number>` versions (where `<number>` is a three-digit, zero padded integer like `-dev003`), if you need it for development. 
```

gspencergoog (approved):

packages/genai_primitives/CHANGELOG.md
polina-c: Yes, you are right.

And this change is samll enough to hold for next release.

Changed.

PR #919, opened 2026-05-14
fix: surface catalog ID in PromptBuilder

gspencergoog:
# Description

This PR Fixes #900 by surfacing the catalog ID in the system prompt generated by `PromptBuilder` when it is provided in the `Catalog`.

## Changes
- **packages/genui/lib/src/facade/prompt_builder.dart**:
  - Added a line to the system prompt: `The active catalog ID is: "${catalog.catalogId}".` if `catalog.catalogId` is not null.
- **packages/genui/test/facade/prompt_builder_test.dart**:
  - Added a test to verify that the catalog ID is surfaced.
  - Removed the test for null fallback as the implementation was changed by the user to not include the line when null.
- **packages/genui/test/facade/prompt_builder_test.golden/**:
  - Updated all 8 golden files to reflect the new prompt content.

## Impact & Risks
- No breaking changes.
- Minimal risk.

## Testing
- Ran `prompt_builder_test.dart` and verified that all 10 tests pass.


packages/genui/lib/src/facade/prompt_builder.dart
```diff
@@ -363,9 +363,15 @@ final class _BasicPromptBuilder extends PromptBuilder {
       catalog,
     ).toJson(indent: '  ');
 
+    String? activeCatalogId = catalog.catalogId
+        ?.replaceAll('"', '\\"')
+        .replaceAll('\n', '\\n');
```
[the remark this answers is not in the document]
ditman: Are we enforcing anywhere that the catalogID is a URL? In that case Gemini's suggestion wouldn't make too much sense, because all those weird characters would end up url-encoded (`%5C`?) and wouldn't break out of the quotes?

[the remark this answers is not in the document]
gspencergoog: Yes, you're right: the catalogID is just a string. I agree, it's not a valid suggestion.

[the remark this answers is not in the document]
gspencergoog: I updated to PR to not sanitize, and modified the description of catalog ID.

packages/genui/lib/src/model/a2ui_schemas.dart
```diff
@@ -436,7 +436,12 @@ abstract final class A2uiSchemas {
         'the component tree.',
     properties: {
       surfaceIdKey: S.string(description: 'The unique ID for the surface.'),
-      'catalogId': S.string(description: 'The URI of the component catalog.'),
+      'catalogId': S.string(
+        description:
+            'A string that uniquely identifies this catalog. This is typically '
+            'a URI, but the URI is never accessed, only used as a unique '
+            'identifier.',
```
ditman: Did this break the tests? The goldens seem to expect a different `description`?

gspencergoog: Yes.  Fixed.

examples/simple_chat/macos/Flutter/GeneratedPluginRegistrant.swift
```diff
@@ -12,5 +12,5 @@ import video_player_avfoundation
 func RegisterGeneratedPlugins(registry: FlutterPluginRegistry) {
   AudioplayersDarwinPlugin.register(with: registry.registrar(forPlugin: "AudioplayersDarwinPlugin"))
   UrlLauncherPlugin.register(with: registry.registrar(forPlugin: "UrlLauncherPlugin"))
-  FVPVideoPlayerPlugin.register(with: registry.registrar(forPlugin: "FVPVideoPlayerPlugin"))
+  VideoPlayerPlugin.register(with: registry.registrar(forPlugin: "VideoPlayerPlugin"))
```
ditman, line 15: Are the changes to the simple_chat and verdure samples related to this change? Or do the examples need rebuilding? Can these be moved to a separate PR?

packages/genui/test/facade/prompt_builder_test.golden/all_operations_with_dataModel_true.txt
```diff
@@ -16,6 +16,10 @@ Use the provided tools to respond to user using rich UI elements.
 
 -------------------------------------
 
+The active catalog ID is: "test_catalog". You must use this catalog ID when creating surfaces.
```
ditman, line 19: I've seen mentions to the "active catalog" in some points of the [spec](https://github.com/a2ui-project/a2ui/blob/852b0624eb067cf6b6aa2eab43600839496acdb1/specification/v1_0/docs/a2ui_protocol.md?plain=1#L368), but they only seem to pertain to the "functions" part of the specification. Is "active catalog" a term that needs to be defined more specifically in the spec?

ditman (approved):
This LGTM, but it'd be nicer if we split the seemingly unrelated changes to the examples and dev_tools to a different PR.

examples/simple_chat/macos/Flutter/GeneratedPluginRegistrant.swift
gspencergoog: These are just generated files that get updated if you run them with the current stable Flutter.  I actually think we shouldn't be checking them in. I'll remove them from this PR and do it in another one.

ditman: I assumed these were automated changes of our sample apps. Thanks for removing them from this PR!

packages/genui/test/facade/prompt_builder_test.golden/all_operations_with_dataModel_true.txt
gspencergoog: Maybe, sure.  I think in context it's fairly obvious here (it's the one that updates will be targeted to).

PR #921, opened 2026-05-14
feat(genui): support functions in prompts and verify rendering

gspencergoog:
## Summary

This pull request enhances the GenUI system to support client-side function declarations registered in the catalog. It updates prompt construction, enables dynamic schema validation via `SchemaRegistry` to support `$ref` references, secures function execution error propagation, and fixes stream pause/resume signal propagation.

### 1. Catalog Function Prompts
- Added function definitions to the LLM system prompt generated by `PromptBuilder` using the expected parameter and return type structure.
- Reused `Catalog.fullSchema` in `PromptBuilder` to unify components schema mapping and keep the schemas DRY.
- Reverted initial async `PromptBuilder` constructor changes back to synchronous constructors (`chat`, `custom`) by reading schemas synchronously from compiled constants.
- Avoided shared schema mutation by copying the `$defs` map before modifying the `anyFunction` property in `PromptBuilder._generateCatalogSchema`.

### 2. Reference Schema Validation
- Refactored component validation in `SurfaceController` to use `SchemaRegistry`, adding `common_types.json` and the dynamically generated `catalog.json` schema.
- Updated `SurfaceDefinition.validate` to be asynchronous to support dynamic schema resolving.
- Pre-populated the fallback `SchemaRegistry` in `SurfaceDefinition.validate` with standard `common_types.json` schema elements to make manual validation calls easier.
- Cached parsed schemas as static final fields in both `SurfaceController` and `SurfaceDefinition` to prevent redundant parsing and json decoding.
- Added null-safe casting for properties in `Catalog.fullSchema` to prevent `TypeErrors` on schemas that do not define properties.

### 3. Structured Function Execution Safety
- Introduced `A2uiFunctionException` to handle client-side function failures with local debugging data (e.g. `cause`).
- Mapped exceptions to the `FUNCTION_EXECUTION_FAILED` protocol response code while omitting the raw exception cause to keep stack traces and internal details hidden from the AI.
- Added a 10-second timeout to the `Button` press execution stream and wrapped errors into user-facing `A2uiFunctionException` wrappers.
- Aligned basic client functions' parameters in `basic_functions.dart` to use dynamic references (e.g. `A2uiSchemas.stringReference()`) so binding path validation passes.

### 4. Stream Extension Correctness
- Updated custom `switchMap` stream helper to propagate pause and resume signals to the outer and inner subscriptions.
- Added regression tests verifying pause/resume propagation to `stream_extensions_test.dart`.

## Impact & Risks

- **Breaking Change**: `SurfaceDefinition.validate` is now asynchronous (`Future<void>`).
- **Breaking Change**: `PromptBuilder` factory constructors have been migrated to static methods.
- **Breaking Change**: Replaced low-level GenUI messages and stores with types from `package:a2ui_core`.

## Testing

- Verified that all static analysis (`flutter analyze --fatal-infos`) passes with zero warnings.
- Ran all `295` unit and widget tests successfully.
- Added regression test `pause and resume are propagated to subscriptions` inside `stream_extensions_test.dart`.
- Added new catalog widget and functions rendering tests in `functions_rendering_test.dart` and `button_test.dart`.


packages/genui/lib/src/facade/prompt_builder.dart
```diff
@@ -359,23 +379,88 @@ final class _BasicPromptBuilder extends PromptBuilder {
...
+        item.name: {
+          'type': 'object',
+          'allOf': [
+            {r'$ref': r'common_types.json#/$defs/ComponentCommon'},
+            {r'$ref': r'#/$defs/CatalogComponentCommon'},
+            {
+              'type': 'object',
+              'properties': {
+                'component': {'const': item.name},
+                ...item.dataSchema.value['properties'] as Map<String, dynamic>,
```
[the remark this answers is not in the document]
gspencergoog: Actually, it does have a value property getter.

packages/genui/lib/src/facade/prompt_builder.dart
```diff
@@ -78,41 +79,55 @@ abstract class PromptBuilder {
   /// The builder will generate a prompt for a chat session,
   /// that instructs to create new surfaces for each response
   /// and restrict surface deletion and updates.
-  factory PromptBuilder.chat({
+  static Future<PromptBuilder> createChat({
```
ditman: This is a breaking change, we need to roll a major version on the genui package (v0.10.0); please also update the CHANGELOG!

packages/genui/lib/src/catalog/basic_catalog_widgets/button.dart
```diff
@@ -233,7 +234,33 @@ Future<void> _handlePress(
     try {
       await resultStream.first;
```
ditman: I know that this is unrelated to this change buuuuttt... Is there any chance this `await stream.first` never finishes?

packages/genui/lib/src/catalog/basic_catalog_widgets/button.dart
```diff
@@ -233,7 +234,33 @@ Future<void> _handlePress(
...
     } catch (exception, stackTrace) {
-      itemContext.reportError(exception, stackTrace);
+      genUiLogger.severe(
+        'Error executing function call "$callName" on button press',
+        exception,
+        stackTrace,
+      );
+
+      if (exception is A2uiFunctionException) {
```
ditman, line 251: Unfortunate that we can't use the `on Type` catching, instead of having all these is checks...

packages/genui/lib/src/catalog/basic_catalog_widgets/date_time_input.dart
```diff
@@ -340,9 +340,7 @@ final dateTimeInput = CatalogItem(
         {
           "id": "root",
           "component": "DateTimeInput",
-          "value": {
-            "path": "/myDateTime"
-          }
+          "value": "2026-05-15"
```
ditman, line 343: Why did we need to change this example?

packages/genui/lib/src/engine/surface_controller.dart
```diff
@@ -138,15 +144,18 @@ interface class SurfaceController implements SurfaceHost, A2uiMessageSink {
         'code': errorCode,
         'surfaceId': ?surfaceId,
         'path': ?path,
+        'functionName': ?functionName,
         'message': message,
       },
     };
```
ditman, line 335: I'm not sure how many more error types we might want to add, but it might make sense to extract this A2ui*Exception -> error map logic to its own separate method (if we don't want to have a `.toJson(version)` method on the Exception implementations themselves)

packages/genui/lib/src/facade/prompt_builder.dart
```diff
@@ -78,41 +79,55 @@ abstract class PromptBuilder {
...
       importancePrefix: importancePrefix,
       clientDataModel: clientDataModel,
       technicalPossibilities: const TechnicalPossibilities(),
+      commonTypesSchema: commonTypes,
+      serverToClientSchema: serverToClient,
     );
   }
 
-  factory PromptBuilder.custom({
+  static Future<PromptBuilder> createCustom({
```
ditman: (This is also a breaking change, make a note in the changelog as well)

packages/genui/lib/src/facade/prompt_builder.dart
```diff
@@ -78,41 +79,55 @@ abstract class PromptBuilder {
...
+      serverToClientSchema: serverToClient,
     );
   }
 
+  static Future<(String, String)> _loadSchemas() async {
+    final String commonTypes = await rootBundle.loadString(commonTypesAssetKey);
+    final String serverToClient = await rootBundle.loadString(
+      serverToClientAssetKey,
+    );
+    return (commonTypes, serverToClient);
```
ditman: I kind of like named records more, instead of using the destructuring assignation on 110, but OK :P (I don't like the `.$n` positional accessors)

packages/genui/lib/src/facade/prompt_builder.dart
```diff
@@ -352,30 +371,147 @@ final class _BasicPromptBuilder extends PromptBuilder {
...
       fragments.map((e) => e.trim());
 
-  final TechnicalPossibilities technicalPossibilities;
-
   @override
   Iterable<String> systemPrompt() {
-    final String a2uiSchema = A2uiMessage.a2uiMessageSchema(
-      catalog,
-    ).toJson(indent: '  ');
+    final String catalogSchema = _generateCatalogSchema(catalog);
```
ditman: Not using the `A2uiMessage` to generate the schema and instead moving the code here to the `_generateCatalogSchema` seems like a step backwards, why do we want it this way?

packages/genui/lib/src/primitives/constants.dart
```diff
@@ -5,3 +5,20 @@
...
+    'packages/genui/assets/schemas/common_types.json';
+
+/// Asset path for server-to-client message envelope schema.
+const String serverToClientAssetKey =
+    'packages/genui/assets/schemas/server_to_client.json';
+
+/// Local filesystem path to common A2UI types schema (for test and development
+/// utilities).
+const String commonTypesLocalPath =
+    'submodules/a2ui/specification/v0_9/json/common_types.json';
```
ditman: It seems that these constants are part of the API ([api docs](https://pub.dev/documentation/genui/0.9.0/genui/#constants)?). Please make sure that we're not increasing the API surface for internal constants that don't need to be exported.

It seems that this file is being exported wholesale (from [here](https://github.com/gspencergoog/genui/blob/0520535faf0b46a614b10a3e9ef37b7378459571/packages/genui/lib/src/primitives.dart#L9) and then [here](https://github.com/gspencergoog/genui/blob/0520535faf0b46a614b10a3e9ef37b7378459571/packages/genui/lib/genui.dart#L21)), but judging by some of the dartdocs these contents are not intended for end-user consumption.

Can we move the ones that are not expected to be consumed by users of the package be exported differntly? Can we add a `show` clause to the `genui.dart` barrel file so we can explicitly control the API surface of the package?

packages/genui/lib/test/validation.dart
```diff
@@ -81,8 +84,27 @@ Future<List<ExampleValidationError>> validateCatalogItemExamples(
       components: components,
     );
 
+    var file = File(
+      'submodules/a2ui/specification/v0_9/json/common_types.json',
```
ditman: Is this path the same as `const commonTypesLocalPath` above?

packages/genui/test/facade/prompt_builder_test.golden/create_only_with_dataModel_true.txt
ditman, line 1: These goldens are getting chonky!

packages/genui/test/facade/prompt_builder_test.dart
```diff
@@ -2,12 +2,53 @@
...
+import 'package:json_schema_builder/json_schema_builder.dart';
 
 import '../test_infra/golden_texts.dart';
 
 void main() {
+  TestWidgetsFlutterBinding.ensureInitialized();
+
+  setUpAll(() {
+    // Mock asset loading because PromptBuilder loads schemas from assets,
+    // and Flutter tests do not load package assets automatically.
```
ditman: I wonder if we could resolve this a little bit more gracefully, like having a way to inject a way to resolve assets that is `@visibleForTesting`, instead of directly using the `rootBundle`?

packages/genui/test/test_infra/validation_test_utils.dart
```diff
@@ -53,8 +56,27 @@ void validateCatalogExamples(
             components: components,
           );
 
+          var file = File(
+            'submodules/a2ui/specification/v0_9/json/common_types.json',
```
ditman: Is this code repeated? I remember the !file.existsSync from a few files above. Can we share this testing logic?

(Also, didn't we add a const with this path?)

packages/genui/assets/schemas/server_to_client.json
ditman, line 1: Diff is weird on this file; is this a symlink to a git submodule?

ditman: QQ, do we need to modify the test runners to clone this package with submodules recursively, or is that the default?

ditman (changes requested):
A few comments on this PR, this introduces a seemingly breaking change (which probably requires a major version release), and also seems to be introducing a bunch of public API that we may not want to expose. The rest of my comments are less critical!

packages/genui/assets/schemas/server_to_client.json
gspencergoog: Yes, the asset files are symbolic links pointing to the spec schemas inside the `submodules/a2ui` directory to keep from duplicating the schemas in multiple places.

When publishing to `pub.dev` with `dart pub publish`, pub automatically resolves the symlinks and bundles the actual file contents, so users of the package will get the actual files and don't need to check out the submodules.

For CI testing, the test runners in `flutter_packages.yaml` are already configured to checkout and update submodules recursively (`submodules: recursive`), so they are there during testing and validation.

packages/genui/lib/src/catalog/basic_catalog_widgets/button.dart
gspencergoog: Yes, if a custom `ClientFunction` returns a stream that never emits any values (e.g., an empty stream), `await resultStream.first` _would_ block. Maybe we should implement a timeout on the future? I added a 10 second timeout, but is that sufficient?  I mean, it could be waiting on a long inference...  Should it be 1 minute? Any user that waits that long is pretty patient. But I suppose it's better than waiting forever.

packages/genui/lib/src/catalog/basic_catalog_widgets/button.dart
gspencergoog: Yeah, I agree.

packages/genui/lib/src/catalog/basic_catalog_widgets/date_time_input.dart
gspencergoog: Because `DateTimeInput` also allows raw string inputs, and I wanted to add an example that covered that.  We already had examples that covered data model values.

packages/genui/lib/src/facade/prompt_builder.dart
gspencergoog: Splitting the schemas with the fenced blocks matches how they are structured in the protocol specification and how the web code does this generation, so that all the platforms benefit from the same optimization through modifying the schema to improve eval results.

It also allows the LLM to use schema references (`$ref`) instead of bundling everything into a single massive, duplicated schema document. This improves schema resolution clarity for the LLM and keeps the prompt size smaller, since it can reuse the `$defs`.

packages/genui/lib/src/facade/prompt_builder.dart
gspencergoog: Good point!  Added to the changelog.

packages/genui/lib/src/engine/surface_controller.dart
gspencergoog: Good idea. Done.

packages/genui/lib/src/facade/prompt_builder.dart
gspencergoog: Good point!  Added to the changelog.

packages/genui/lib/src/facade/prompt_builder.dart
gspencergoog: Sure, converted.

packages/genui/lib/src/primitives/constants.dart
gspencergoog: There were several things exported by `primitives.dart` that didn't need to be exported (like the entire `constants.dart` contents).  I added show clauses to `primitives.dart` to limit what gets exported.  Also added that to the `CHANGELOG.md` as a breaking change, since someone could theoretically been depending on it.

packages/genui/lib/test/validation.dart
gspencergoog: Yes, consolidated.

packages/genui/test/facade/prompt_builder_test.golden/create_only_with_dataModel_true.txt
gspencergoog: Yes, yes they are.

packages/genui/test/facade/prompt_builder_test.dart
gspencergoog: To avoid duplicating this mocking boilerplate in other test files, I extracted the mock binary messenger setup into a shared helper function under `test_infra/` so it can be reused.

packages/genui/test/test_infra/validation_test_utils.dart
gspencergoog: Yes, consolidated.

ditman (approved):
I don't have much more to say about this PR! If this looks good to people who're working more with the flutter codebase (maybe @andrewkolos?) we can ship and maybe publish `genui` and `genui_a2a`! :)

packages/genui/lib/src/model/a2ui_schemas.dart
```diff
@@ -383,46 +374,17 @@ abstract final class A2uiSchemas {
...
-      required: ['event'],
-    );
-
-    final functionCallSchema = S.object(
-      properties: {'functionCall': functionCall()},
-      required: ['functionCall'],
-    );
-
     return S.combined(
+      $ref: '$_commonTypesUri#/\$defs/Action',
```
andrewkolos, line 378: I don't see any code in schema_validation.dart that follows `$ref`s. If this is the case, than we lose runtime validation for any properties we define using refs (like `action` and `validationCheck`; a bunch of other ones are affected too but would also need #955 merged to work)

gspencergoog:
CC @redbrogdon You might be interested in this for your talk.

packages/genui/lib/src/model/a2ui_schemas.dart
gspencergoog: So, after #955 merges, would this still be validated?

gspencergoog: How about this ( https://github.com/flutter/genui/pull/998) as a fix for the `$ref` validation?  I think we should do a full validation using `json_schema_builder`, and not just use a simple approximation.  Does that make sense?

gspencergoog: Okay, so after the #998 PR, the `$ref` validation is hardened, and I've merged that into this PR, so I'm going to go ahead and commit this.

packages/genui/lib/src/facade/prompt_builder.dart
```diff
@@ -372,13 +401,47 @@ final class _BasicPromptBuilder extends PromptBuilder {
...
+      };
+    } else {
+      catalogJson.remove('functions');
+      final defs = catalogJson[r'$defs'] as Map<String, dynamic>;
+      defs['anyFunction'] = {'not': <String, dynamic>{}};
+    }
+
+    final String json = const JsonEncoder.withIndent('  ').convert(catalogJson);
+    return json.replaceAll(commonTypesSchemaId, 'common_types.json');
+  }
```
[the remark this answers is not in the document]
gspencergoog: Done. Copied the $defs map before modifying the `anyFunction` definition to keep the original catalog schema immutable.

packages/genui/lib/src/utils/stream_extensions.dart
```diff
@@ -17,3 +18,57 @@ extension CombineLatestAll<T> on Iterable<Stream<T>> {
...
+            }
+          },
+        );
+      },
+      onCancel: () {
+        cancelInner();
+        outerSubscription?.cancel();
+        outerSubscription = null;
+      },
+    );
```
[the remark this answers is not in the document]
gspencergoog: Done. Propagated stream pause and resume events to the underlying subscriptions inside the custom switchLatest implementation.

packages/genui/lib/src/engine/surface_controller.dart
```diff
@@ -57,16 +58,16 @@ interface class SurfaceController implements SurfaceHost, A2uiMessageSink {
...
       Uri.parse(commonTypesSchemaId),
       Schema.fromMap(jsonDecode(commonTypesSchemaJson) as Map<String, Object?>),
     );
-    _schemaRegistry = registry;
+    registry.addSchema(
+      Uri.parse('https://a2ui.org/specification/v0_9/catalog.json'),
+      catalog.fullSchema,
+    );
     return registry;
   }
```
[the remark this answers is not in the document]
gspencergoog: Done. Cached the parsed common types schema as a static final field to avoid parsing JSON repeatedly.

packages/genui/lib/src/model/ui_models.dart
```diff
@@ -169,13 +170,22 @@ class SurfaceDefinition {
...
   Future<void> validate(Schema schema, {SchemaRegistry? registry}) async {
+    final SchemaRegistry reg = registry ?? SchemaRegistry();
+    if (registry == null) {
+      reg.addSchema(
+        Uri.parse(commonTypesSchemaId),
+        Schema.fromMap(
+          jsonDecode(commonTypesSchemaJson) as Map<String, Object?>,
+        ),
+      );
+    }
```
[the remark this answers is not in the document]
gspencergoog: Done. Cached the parsed common types schema as a static final field in SurfaceDefinition.

packages/genui/lib/src/model/catalog.dart
```diff
@@ -244,6 +245,120 @@ interface class Catalog {
...
+        item.name: {
+          'type': 'object',
+          'allOf': [
+            {r'$ref': '$commonTypesSchemaId#/\$defs/ComponentCommon'},
+            {r'$ref': r'#/$defs/CatalogComponentCommon'},
+            {
+              'type': 'object',
+              'properties': {
+                'component': {'const': item.name},
+                ...item.dataSchema.value['properties'] as Map<String, dynamic>,
```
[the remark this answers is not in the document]
gspencergoog: Done. Used a null-safe cast with a fallback empty map for properties to prevent TypeError when properties are missing.

PR #925, opened 2026-05-15
feat(ci): enforce monorepo test coverage thresholds and baseline tracking

gspencergoog:
# Description

This PR adds automated test coverage threshold checking and baseline regression tracking to the monorepo.

When tests run in CI or via local tooling, a verification step parses the resulting `lcov.info` trace files and compares the actual coverage percentage against minimum thresholds and previous high-water marks. If a package's coverage regresses below its baseline, the check fails.

Key components:
* `coverage_policy.yaml`: Root configuration file defining global file exclusions (such as generated code) and package-specific minimum thresholds. Path traversal checks are enforced to prevent baseline files from pointing outside the repository.
* `coverage_baseline.yaml`: Tracking file recording high-water coverage percentages for all tested packages. A tolerance of 0.2% is allowed before failing on regression to account for minor platform differences.
* `tool/test_and_fix`: Updated to accept `--coverage` and `--update-baseline` flags. Added automated untested file accounting by generating temporary full-coverage test files before execution. It uses an AST parser to skip `part of` files during generation. Trace file parsing is delegated to `package:lcov_parser`. When running in coverage mode, global fixes, formatting, and analysis jobs are skipped to speed up CI. Submodules are also excluded from discovery. `exitCode` is explicitly set to report failures correctly to the Dart VM.
* `.github/workflows`: Added a dedicated `coverage` verification job to `flutter_packages.yaml` to run on all pushes and pull requests to `main`. The step checks out submodules recursively to support packages that depend on them for testing.

Added comprehensive unit tests for the coverage engine, achieving over 94% test coverage for the `test_and_fix` package.

Fixes https://github.com/flutter/genui/issues/920


gspencergoog:
Okay, this adds coverage testing to the repo.  One thing to notice, though, is that the test_and_fix program that I'm doing the coverage parsing/collection in is also running all of the tests.  It runs them in parallel, so it is as efficient as running all of the separate jobs.  I'm wondering if we can eliminate all of the separate jobs and just run the tests once, and collect both coverage and testing information at the same time.

If we still want separate github actions for some reason, we could modify the program to just do coverage and testing on the directory given in an argument.

polina-c (approved):
LGTM

Thank you for adding this!

PR #926, opened 2026-05-16
Move submodules to repo root.

polina-c:
Contributes to https://github.com/google/A2UI/issues/1383.

Prerequisite for https://github.com/flutter/genui/pull/917

Fix the issue wuth submodules by moving them to root.

ISSUE: 

`flutter pub publish` gives a warning that gitignored files are committed.

Pub's gitignore validator reports submodules/a2ui as ignored even though no .gitignore rule actually matches it — git check-ignore returns nothing locally. So pub uses its own logic for this, not real git ignore semantics, and it appears to treat an uninitialized submodule gitlink as "ignored."

### DIFF
--- a/.agent/skills/genui-helper/SKILL.md
+++ b/.agent/skills/genui-helper/SKILL.md
@@ -51,9 +51,9 @@ When creating a new UI component in `genui`:
 ## References
 
 - A2UI Specification
-  - Available in the submodule at @packages/genui/submodules/a2ui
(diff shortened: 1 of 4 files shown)


andrewkolos (approved):
RSLGTM, though people with existing checkouts will probably have to resync submodules to avoid orphaned files (and broken tests if i had to guess)

edit: this should do the job:

```
git submodule update --init --recursive
```

PR #937, opened 2026-05-27
fix(genui): A2uiMessage.fromJson rejects messages with more than one type

andrewkolos:
Fixes #934.

Related: https://github.com/flutter/genui/pull/936. fixes the same bug in a2ui_core


andrewkolos:
Closing. I think this can wait until `genui` is rebuilt on top of `a2ui_core`, where #936 will fix

PR #938, opened 2026-05-27
fix(a2ui_core): distinguish between `"key": null` and absence of `key` in DataModel

andrewkolos:
See https://github.com/flutter/genui/issues/935.

This bug is also present in `genui`. However, to reduce churn, I recommend just waiting until we rebuild `genui` on top of `a2ui_core`.

### DIFF
--- a/packages/a2ui_core/lib/src/core/data_model.dart
+++ b/packages/a2ui_core/lib/src/core/data_model.dart
@@ -94,13 +94,13 @@ class DataModel {
     Object? current = _data;
     for (var i = 0; i < dataPath.segments.length - 1; i++) {
       final String segment = dataPath.segments[i];
-      final String nextSegment = dataPath.segments[i + 1];
-      final isNextNumeric = int.tryParse(nextSegment) != null;
 
       if (current is Map<String, Object?>) {
         if (!current.containsKey(segment) || current[segment] == null) {
           if (!autoVivify) return null;
-          current[segment] = isNextNumeric ? <Object?>[] : <String, Object?>{};
+          current[segment] = int.tryParse(dataPath.segments[i + 1]) != null
+              ? <Object?>[]
+              : <String, Object?>{};
         }
         current = current[segment];
       } else if (current is List<Object?>) {
@@ -124,7 +124,9 @@ class DataModel {
           while (current.length <= index) {
             current.add(null);
           }
-          current[index] = isNextNumeric ? <Object?>[] : <String, Object?>{};
+          current[index] = int.tryParse(dataPath.segments[i + 1]) != null
+              ? <Object?>[]
+              : <String, Object?>{};
         }
         current = current[index];
       } else {
--- a/packages/a2ui_core/lib/src/core/data_model.dart
+++ b/packages/a2ui_core/lib/src/core/data_model.dart
@@ -106,12 +106,14 @@ class DataModel {
       } else if (current is List<Object?>) {
         final int? index = int.tryParse(segment);
         if (index == null) {
+          if (!autoVivify) return null;
           throw A2uiDataError(
             "Cannot use non-numeric segment '$segment' on a list.",
             path: path,
           );
         }
         if (index < 0 || index > maxAutoVivifyIndex) {
+          if (!autoVivify) return null;
           throw A2uiDataError(
             'List index out of bounds: $index (max $maxAutoVivifyIndex)',
             path: path,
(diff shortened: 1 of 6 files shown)


packages/a2ui_core/lib/src/core/data_model.dart
```diff
@@ -43,90 +43,127 @@ class DataModel {
...
+          );
+        }
+        if (index >= current.length || current[index] == null) {
+          if (!autoVivify) return null;
           while (current.length <= index) {
             current.add(null);
           }
-          current[index] = value;
+          current[index] = isNextNumeric ? <Object?>[] : <String, Object?>{};
         }
```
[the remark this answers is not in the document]
andrewkolos: Addressed.

packages/a2ui_core/lib/src/core/data_model.dart
```diff
@@ -43,90 +43,127 @@ class DataModel {
...
+        throw A2uiDataError(
+          'List index out of bounds: $index (max $maxAutoVivifyIndex)',
+          path: path,
+        );
+      }
+      while (parent.length <= index) {
+        parent.add(null);
+      }
+      parent[index] = value;
+    }
```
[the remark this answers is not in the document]
andrewkolos: Addressed.

packages/a2ui_core/lib/src/core/data_model.dart
```diff
@@ -43,90 +43,136 @@ class DataModel {
...
-        }
+        final Object? parent = _walkToParent(dataPath, path, autoVivify: true);
+        _writeLeaf(parent, dataPath.segments.last, value, path);
+      }
+      _notifyPathAndRelated(dataPath);
+    });
+  }
+
+  /// Removes the value at [path]. For map keys this deletes the key; for
+  /// list indices it sets the slot to null and preserves length.
```
gspencergoog, line 63: When are the nulls in lists cleaned up?  And how are they handled when rendering?  Wouldn't you expect the list to change size if you remove an entry?

andrewkolos: Great questions that I didn't think much about since I was really only thinking about conforming to the current a2ui renderer guide (as best could be done).

> When are the nulls in lists cleaned up?

They aren't. For all intents and purposes, the server would need to replace the list with a new, compacted list. This PR treats list removal as preserving length and setting the slot to `null` ([data_model.dart#L62-L81](https://github.com/flutter/genui/blob/9b226c5df0b57b432444d11f63a0786b651b91f8/packages/a2ui_core/lib/src/core/data_model.dart#L62-L81)).

> How are they handled when rendering?

They are still iterated over. `genui` does not filter null entries during list-template iteration: `ComponentChildrenBuilder` delegates template-list rendering to the catalog widget ([widget_helpers.dart](https://github.com/flutter/genui/blob/9b226c5df0b57b432444d11f63a0786b651b91f8/packages/genui/lib/src/catalog/basic_catalog_widgets/widget_helpers.dart#L47)), and `List` builds one child per list index ([list.dart#L99-L124](https://github.com/flutter/genui/blob/9b226c5df0b57b432444d11f63a0786b651b91f8/packages/genui/lib/src/catalog/basic_catalog_widgets/list.dart#L99-L124)). `a2ui_core` does the same thing: the binder generates a `ChildNode` for each list index with the corresponding scoped base path ([binder.dart#L124-L133](https://github.com/flutter/genui/blob/9b226c5df0b57b432444d11f63a0786b651b91f8/packages/a2ui_core/lib/src/rendering/binder.dart#L124-L133)).

This is also what the v0.9 renderer guide specifies. It says setting an array index to `undefined` preserves length and empties the index ([renderer_guide.md#L284](https://github.com/google/A2UI/blob/305336f137034922053874471d95614a3831e186/specification/v0_9/docs/renderer_guide.md?plain=1#L284)). Dart/JSON do not have JS-style `undefined`, and JSON cannot represent sparse array slots. `web_core` can keep `undefined` in-memory. In `a2ui_core`, the equivalent emptied array slot is represented as `null` (I suppose we could use maps and represent empty list slots that way...but yeah that's intuitive complexity for behavior I am not even 100% sure is worth preserving).

This whole situation is a bit complex and not ideal, but I'll get to that in my next point.

> Wouldn't you expect the list to change size if you remove an entry?

Intuitively, yes, but that's not under the current v0.9 semantics. “Remove object key” and “empty array slot” are different operations. If we want remove-from-list to mean splice/compact the list, I think that should be a spec/API change rather than part of this PR. 

I can file an issue over in the A2UI repo (as well as one here) to track work to make the answers to your questions intuitive. WDYT?

andrewkolos: Filed https://github.com/google/A2UI/issues/1504

andrewkolos:
Closing as blocked on https://github.com/google/A2UI/issues/1504

PR #948, opened 2026-05-31
Add recommendations for hosting binaries.

polina-c:


polina-c:
Helping video, linked in the recommendations:


https://github.com/user-attachments/assets/862e4f6a-ed11-4fa9-b014-c92b19d8aca2

docs/contributing/README.md
```diff
@@ -24,9 +24,20 @@ I have read and understood ./docs/contributing/README.md.
...
 ## Binary files
 
-Avoid adding binary files to the repository. If a binary file is needed, minimize its size and accompany it with a markdown file that 
+Avoid adding binary files to the repository. If a binary file is needed, minimize its size
+and accompany it with a markdown file that 
 describes the binary file and how it was created.
 
+For large binaries, use links instead of embedding them in the repository.
+
+To host binaries, upload them to your personal account on a public service:
```
ditman, line 33: I don't think this is good advice, you probably don't want to host an asset for a demo app on google drive, flickr or a github issue comment (they all seem very easy to bit-rot).

If a demo app *requires* big binaries, it probably should live in a separate repo. If we want, we could provide an `a2ui-org/assets` repo so big static assets can be provided from there (as a submodule, for example)?

yjbanov: I agree that we should not offer this path. I think for large binaries we should have a serious discussion for why the binary is needed, how to host it, how to maintain it, etc. The default answer for large binaries should be "no, don't add it".

PR #963, opened 2026-06-10
Update publishing to pub.dev

polina-c:
Add instructions for on-call.


docs/contributing/publishing.md
```diff
@@ -70,6 +70,28 @@ major number for breaking changes.
...
+   
+   If publishable changes are under `wip` in `CHANGELOG.md`:
+
+   2.1. Convert the wip version to a non-wip version in `CHANGELOG.md`.
+
+   2.2. Publish the new version.
+   
+   2.3. Comment on the PR that merged publishable changes under `wip`:
+    
+     > This PR wrongly placed publishable changes under a `wip` version. This is corrected in PR <link>. See guidelines [here](https://github.com/flutter/genui/blob/main/docs/contributing/publishing.md#-wip-vs-non--wip-production-ready-versions).
```
ditman, line 95: Needing this makes me feel that there's a point in the process that introduces ambiguity when reviewing. Do you think the "-wip" should be just a "escape hatch" when a change really really does *not* need to be published, that the reviewers can suggest to do?

For example, in flutter/packages, there's the same expectation of "auto-publish after every push", but what this flow calls "-wip" is called "NEXT" there (see [this](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changelog)). NEXT changes are not expected to be published at all (and the automation skips them), and only when a real release happens, their changelog entries get merged into that of the release.

ditman (approved):
I think having an explicit "-wip" choice makes this process a little bit more complicated; but I guess that's the Dart standard so let's go!

docs/contributing/publishing.md
polina-c: Interesting.

I like the direction.

For me "NEXT" also means some changes. 

I like 'noop' more.

And, 'wip' versions still may exist as development artifact, needed for testing.

I will review the doc from this perspective.

docs/contributing/publishing.md
```diff
@@ -19,8 +19,8 @@ To make sure your PR passes this validation, follow [firehose rules](https://git
 
 Packages in this repo fall into the following categories:
 
-1. **Not published**: `pubspec.yaml` contains `publish_to: none`. Workspace tools and example apps that are never pushed to pub.dev.
-2. **Not yet published**: the package's `version:` ends with a `-wip<N>` suffix (see "`-wip` vs non-`-wip`" below). Published to pub.dev to reserve the name or to test the package; not ready for general use yet.
+1. **Not planned to be published**: `pubspec.yaml` contains `publish_to: none`. Workspace tools and example apps that are never pushed to pub.dev.
+2. **Not yet published**: the package's `version:` ends with a `-wip<N>` suffix (see "Versioning" below). Not-ready-for-production versions are pushed to pub.dev to reserve the name and maybe to try the package in dev purposes.
```
ditman, line 23: Reserving the name only matters the first time that the package is published (and it could be version 0.0.0), I don't think we'll ever publish a -wip version to reserve a name (but yes for dev purposes)

docs/contributing/publishing.md
```diff
@@ -60,6 +46,24 @@ major number for breaking changes.
 
 [Semver]: https://semver.org/ 
 
+The versions may have postfixes:
+
+- **`-wip<three digit number>`**: not ready for production
+- **`-noop`**: used in CHANGELOG.md and pubspec.yaml to indicate that the code does not contain publishable changes comparing to the previously published version and thus should not be published to pub.dev.
```
ditman, line 50: I think that instead of a `-noop` suffix (that changes the effective version of the package), changes that don't need to be published should just not upgrade the version of the package (and maybe have an "Unpublished" section in the CHANGELOG if they ever need a version). That way the tooling that we have, will not even let us publish the version (because it'd be "already published")

`-wip` to release things for "integration" testing or pre-release validation, a full normal version (which should be the same as "removing `-wip`") to fully publish the package in a way that would get picked up by pub get.

(We do the "Unreleased" thing in the npm packages already; keeping track of unreleased changes and making a note there if they're breaking/non-breaking in the changelog makes picking the next correct version very easy later)

ditman (approved):
LGTM! Thanks for expanding on the usage of `-wip` and `-noop`!

docs/contributing/publishing.md
polina-c: Thanks.

We will need unreleased/noop in order to be compliant with automation, that verifies version in md is intact with pub.dev and pubspec.yaml.

I prefer 'noop' here, because we want 'op' changes to be released immediately after merge.

.github/workflows/pub_health.yaml
```diff
@@ -0,0 +1,21 @@
...
+# to the org termissions, to enable this workflow
+
+name: Health
+on:
+  pull_request:
+    branches: [ main ]
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+jobs:
+  health:
+    uses: dart-lang/ecosystem/.github/workflows/health.yaml@main
```
ditman, line 17: Should this be pinned at a specific revision of the workflow, and not let it be `@main`?

(Also it's a little bit weird that we're depending on actions directly like this, instead of on something that has been "published", are we allowed to take this dependency from outside dart-lang? (I'm assuming yes)

docs/contributing/publishing.md
ditman: I don't understand the need for `noop`:

* The first time you push a "noop" change, it makes the package publishable, so I may end up with a "-noop" version on pub. (This is probably not what I want)
* Other "noop" changes result in non-publishable changes (this is wanted)

Wouldn't it be easier for "noop" changes to not change the pubspec.yaml at all, so they're effectively non-publishable (because we can't re-publish a version that already exists?)

ditman (commented):
Let's talk offline about the noop!

.github/workflows/pub_health.yaml
polina-c: I am working with flutter org owners to explicitely allow these workflows.

See https://github.com/flutter/genui/blob/main/docs/contributing/publishing.md#setup-org-permissions

So, no, we do not need version

And, yes, everything is secure enough

polina-c:
Yes, this is an issue, and there is a bug for it: https://github.com/dart-lang/ecosystem/issues/421

PR #972, opened 2026-06-15
fix(ci): skip E2E tests when REPO_GEMINI_API_KEY is not set

gspencergoog:
## Summary

Only run the E2E test job in the eval workflow if the `REPO_GEMINI_API_KEY` secret is not empty. This avoids workflow failures on branch pushes from Dependabot, which does not have access to repository secrets.

## Changes

*   Modified the `eval-job` conditional in `.github/workflows/e2e.yaml` to include `secrets.REPO_GEMINI_API_KEY != ''`.

## Impact & risks

There are no risks. On Dependabot runs, the E2E job will be skipped, which satisfies required status checks. It will continue to run as expected on collaborator pushes and scheduled runs.

## Testing

*   Checked that the conditional syntax is valid.


.github/workflows/e2e.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
   eval-job:
     # Do not run on forked branches,
     # because the test does not have access to secrets in forks.
-    if: github.repository == 'flutter/genui'
+    if: github.repository == 'flutter/genui' && secrets.REPO_GEMINI_API_KEY != ''
```
polina-c: How about leaving just second condition so that forks can have key configured and test running?

polina-c (approved):
LGTM with one comment

.github/workflows/e2e.yaml
gspencergoog: Good point! I've updated the condition to only check `secrets.REPO_GEMINI_API_KEY != ''` so that forks can run the tests if they have the key configured.

PR #974, opened 2026-06-16
Migrate `package:genui` onto `package:a2ui_core` (state engine + messages)

andrewkolos:
Advances #811. Supersedes #956.

Migrates `package:genui`'s runtime onto the shared, pure-Dart `package:a2ui_core`: protocol parsing, message processing, and surface/component/data-model state. `SurfaceController` becomes a thin wrapper over `a2ui_core.MessageProcessor`; genui no longer maintains its own parallel A2UI runtime.

The catalog-widget authoring API is unchanged; only the A2UI **message** types move onto `a2ui_core`. Unifying the catalog/component types with the core models is a separate follow-up (#801) that also changes the authoring API.

## Breaking changes

- **The `a2ui_core` message types are now used over genui's bespoke ones.** The genui message classes (`A2uiMessage`, `CreateSurface`, `UpdateComponents`, `UpdateDataModel`, `DeleteSurface`) are removed. `SurfaceController.handleMessage`, `Transport.incomingMessages`, and `A2uiMessageEvent.message` use `core.A2uiMessage`. Construct with `core.CreateSurfaceMessage(...)` etc.; `core.UpdateComponentsMessage` carries raw component JSON maps. Message-handling consumers should depend on `a2ui_core` directly.
- **`SurfaceController.store` / `DataModelStore` removed.** Access a surface's data model via `SurfaceController.contextFor(id).dataModel`.
- **`SurfaceRegistry.updateSurface(...)` removed.** Surface lifecycle flows through `SurfaceController.handleMessage`.

## Behavioral changes inherited from `a2ui_core`

- `DataModel` writes are stricter (errors on type-mismatched intermediate paths and very large list indices; sparse list writes fill skipped entries with `null`).
- A duplicate `createSurface` for an active surface id is now an error.

Full list in packages/genui/CHANGELOG.md. Consumer-facing details migration_0.9.1_to_0.10.0.md.

## Note for reviewers

The significant changes are found across less than 10 files. Everything else is mechanical renames and tests.
Some files of note:

- `packages/genui/lib/src/engine/surface_controller.dart`: `MessageProcessor `delegation, `handleMessage(core.A2uiMessage)`, and buffering of messages that arrive before their target surface is created.
- `packages/genui/lib/src/model/data_model.dart`: `InMemoryDataModel` over `core.DataModel`; `_SignalNotifier` exposes core signals as `ValueListenable`s.
- `packages/genui/lib/src/model/ui_models.dart`: `SurfaceDefinition` / `Component` snapshots; `SurfaceUpdate` events.
- `packages/genui/lib/src/engine/surface_registry.dart`: surface lifecycle.
- `packages/genui/lib/src/widgets/surface.dart`: definition-snapshot render path.
- `packages/genui/lib/src/model/a2ui_message.dart`: now just the catalog-parameterized message schema (the message classes are gone).
- `packages/genui/lib/src/model/schema_validation.dart`, `widget_utilities.dart`.


packages/genui/lib/src/model/schema_validation.dart
```diff
@@ -0,0 +1,174 @@
...
+    'properties': {'component': Map<String, Object?> compProp},
+  }) {
+    return switch (compProp) {
+      {'const': String constType} when constType == type => true,
+      {'enum': List<Object?> enums} when enums.contains(type) => true,
+      _ => false,
+    };
+  }
+  return false;
+}
```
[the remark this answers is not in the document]
andrewkolos: A `Map<String,dynamic>` is a `Map<String,Object?>`, no?

packages/genui/lib/src/model/schema_validation.dart
```diff
@@ -0,0 +1,174 @@
...
+/// Throws [A2uiValidationException] on the first component that fails. State
+/// is not rolled back.
+void validateComponents({
+  required String surfaceId,
+  required Iterable<({String id, String type, JsonMap json})> components,
+  required Schema schema,
+}) {
+  final List<Map<String, Object?>> allowedSchemas = _extractAllowedSchemas(
+    jsonDecode(schema.toJson()) as Map<String, Object?>,
+  );
```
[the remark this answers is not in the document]
andrewkolos: I don't believe the statement about the type error is accurate.

packages/genui/lib/src/engine/surface_controller.dart
```diff
@@ -87,37 +89,178 @@ interface class SurfaceController implements SurfaceHost, A2uiMessageSink {
...
+    // Migrate pre-create fallback data into the live model BEFORE notifying
+    // registry listeners; otherwise a synchronous listener could call
+    // contextFor(...).dataModel and cache an empty live wrapper before the
+    // fallback's data is copied in.
+    final DataModel live = InMemoryDataModel.wrap(surface.dataModel);
+    final DataModel? fallback = _preCreateDataModels.remove(surface.id);
+    if (fallback != null) {
+      live.update(DataPath.root, fallback.getValue<Object?>(DataPath.root));
+      fallback.dispose();
+    }
```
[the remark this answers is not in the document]
andrewkolos: `CreateSurfaceMessage` carries no initial data, so the live model is empty at creation, there's nothing to overwrite. the `!= null` guard wouldn't even fire. An unwritten I`nMemoryDataModel`'s `getValue(root)` is `{}`, not `null`.

examples/simple_chat/lib/a2ui_transport.dart
```diff
@@ -4,6 +4,7 @@
 
 import 'dart:async';
 
+import 'package:a2ui_core/a2ui_core.dart' as core;
```
jacobsimionato, line 7: What are the symbol conflicts that prevent importing both libraries? I think it's worth us making sure that they don't have conflicting symbols in the longer run. If there's some easy way to avoid it now, that would be nice to, avoid all our clients adding `as core` throughout their codebase then needing to remove it later.

packages/genui/lib/src/model/catalog.dart
```diff
@@ -262,3 +268,26 @@ class CatalogItemNotFoundException implements Exception {
...
+
+  @override
+  Schema get schema => _item.dataSchema;
+}
+
+/// Builds the `a2ui_core` [core.Catalog] for [catalog], used when constructing
+/// a [core.SurfaceModel] so `a2ui_core` lookups see real component metadata
+/// instead of an empty stub.
+@internal
+core.Catalog<core.ComponentApi> coreCatalogFor(Catalog catalog) =>
```
jacobsimionato, line 289: Ahhh this is where the symbol collisions happen! I wonder if there is something we can do here to avoid developers needing to add `core.` everywhere in their codebases temporarily, and the removing it when we resolve the conflicts.

I can't think of a better solution than what you've done though. Could we make `Catalog` implement `core.Catalog<core.ComponentApi>` somehow, rather than needing it to be converted?

packages/genui/lib/src/widgets/surface.dart
jacobsimionato, line 1: Is the plan to later refactor this to use `SurfaceModel` instead? I wonder if it could have an alternative constructor or something, so we can start checking in the new logic now without breaking things?

jacobsimionato (commented):
Hey overall, this looks miraculously non-disruptive for developers to me! Nice work!

My main question is whether you have any ideas about avoiding developers needing to add the `core.` import prefixes to resolve symbol conflicts temporarily. Probably there's nothing we can do, but it'd be amazing if we could avoid it. Could we have a separate top-level import file for the conflicting symbols perhaps?

I didn't hit LGTM yet just because I ran out of time and want to spend a bit more tomorrow.

examples/simple_chat/lib/a2ui_transport.dart
andrewkolos: I think you've since found them, but for future reference, the collisions are: `Catalog`, `DataModel`, `DataPath`, `DataContext`, `CancellationException`, `CancellationSignal`, `ExpressionParser`, and `FormatStringFunction`

andrewkolos:
> My main question is whether you have any ideas about avoiding developers needing to add the `core.` import prefixes to resolve symbol conflicts temporarily. Probably there's nothing we can do, but it'd be amazing if we could avoid it. 

I didn't consider this exhaustively. I pretty much always default to the aliasing `import/as` syntax whenever name collisions are involved`–it's more robust when working on frequently changing code and reduces ambiguity when reading code when the reader knows there are many name collisions in the current context[^1].

> Could we have a separate top-level import file for the conflicting symbols perhaps?

Unless I'm missing something, isn't this equivalent to consumers switching to the selective `show` import syntax? Regardless, if you have a preference for `show` over `as`, I'm happy to reconsider more thoughtfully.

[^1]: Aside : a classic example from my personal experience is from game programming where libraries for physics, 3d rendering, computer vision, spatial audio, etc. all export primitive math types/functions like `Vector3` etc.

packages/genui/lib/src/widgets/surface.dart
jacobsimionato, line 28: Not for this PR: A good next step could be to update this so that it can accept a core SurfaceModel and render based on that, doing whatever conversions are necessary. I think the SurfaceModel should expose the necessary data, and allow us to delete SurfaceRegistry etc.

packages/genui/lib/src/engine/surface_controller.dart
```diff
@@ -5,70 +5,69 @@
...
 import '../primitives/logging.dart';
 
-import 'data_model_store.dart';
 import 'surface_registry.dart' as surface_reg;
 
 /// The runtime controller for the GenUI system.
 ///
-/// Orchestrates the lifecycle of UI surfaces, manages communication with the
-/// AI service, and handles data model updates.
+/// Wraps [core.MessageProcessor] and adds Flutter-side concerns: holding
```
jacobsimionato, line 31: BTW all this extra Flutter-specific logic to handle missing catalogs, and cache messages for surfaces that don't exist isn't necessary in the long run. If we want to add this functionality, we can add it to the unified core library design, or more likely to agent SDKs, to ensure they fix payloads before being sent to the client.

But I think it's awesome that this avoids breaking changes for now. I think we can remove this class later, once we add the node API and are ready to complete the migration.

packages/genui/lib/src/model/catalog.dart
```diff
@@ -262,3 +270,26 @@ class CatalogItemNotFoundException implements Exception {
     return buffer.toString();
   }
 }
+
+class _CatalogItemComponentApi implements core.ComponentApi {
```
jacobsimionato, line 274: I'm trying to think of how we can advance aspects of this towards the unified design as fast as possible with minimal confusion. What do you think about just making the genui CatalogItem and Catalog just inherit from the Core models directly, so no conversion is necessary?

This could be a followup PR, if you agree with the approach. Here's a proposal from Gemini - not sure if it has holes:

# Design Document: Catalog Unification & Signal-Based Reactivity

## 1. Goal

Unify the `genui` and `a2ui_core` Catalog and Client Function APIs. By aligning `genui` components and functions directly with `a2ui_core` interfaces, we can:
- Eliminate the translational/adapter layers (`coreCatalogFor`, `_CatalogItemComponentApi`).
- Replace complex, resource-heavy Stream-based reactivity with lightweight, synchronous, and Signal-based updates using `preact_signals`.
- Remove the redundant `ExpressionParser` in `genui` in favor of `a2ui_core`'s native Expression and Data Pointer resolution.

---

## 2. Current Architecture & Bottlenecks

### Component API Redundancy
`genui` uses `CatalogItem` to define a component, which is then mapped at runtime to `core.ComponentApi` via a helper `_CatalogItemComponentApi` inside `coreCatalogFor`. This translation happens during every surface creation.

### Stream vs. Signal Impedance Mismatch
`genui.ClientFunction` is asynchronous and returns a `Stream<Object?>`:
```dart
abstract interface class ClientFunction {
  Stream<Object?> execute(JsonMap args, ExecutionContext context);
}
```
However, the underlying `a2ui_core` engine is built on **Signals** (`preact_signals`). This forces `genui` to:
1. Wrap synchronous functions (e.g., `and`, `or`, `regex`, `length`) in `Stream.value(...)`.
2. Maintain a complex, recursive Stream-based `ExpressionParser` in `genui` to combine and evaluate multiple stream arguments.
3. Convert streams back to signals/notifiers at the widget rendering layer (e.g., `BoundString`).

---

## 3. Proposed Unified Architecture

```
                 +-----------------------+
                 |   core.ComponentApi   |
                 +-----------+-----------+
                             ^
                             | (Implements)
                 +-----------+-----------+
                 |   genui.CatalogItem   |
                 +-----------------------+

             +-------------------------------+
             |  core.FunctionImplementation  |
             +---------------+---------------+
                             ^
                             | (Implements)
             +---------------+---------------+
             |     genui.ClientFunction      |
             +-------------------------------+
```

### A. Unified Component API
`CatalogItem` will directly implement `core.ComponentApi`:
```dart
class CatalogItem implements core.ComponentApi {
  @override
  final String name;

  @override
  Schema get schema => dataSchema;

  final CatalogWidgetBuilder widgetBuilder;
  final bool isImplicitlyFlexible;
  final List<ExampleBuilderCallback> exampleData;
}
```

### B. Signal-Based Client Functions
`ClientFunction` will directly implement `core.FunctionImplementation`, returning either a static value or a `ReadonlySignal<Object?>`:
```dart
abstract class ClientFunction implements core.FunctionImplementation {
  @override
  String get name;

  @override
  String get description;

  @override
  Schema get argumentSchema;

  @override
  core.A2uiReturnType get returnType;

  /// Executes synchronously. Returns a raw value, a ReadonlySignal, or null.
  @override
  Object? execute(
    Map<String, dynamic> args,
    core.DataContext context, [
    core.CancellationSignal? cancellationSignal,
  ]);
}
```

### C. Direct Catalog Inheritance
`genui.Catalog` directly extends `core.Catalog<CatalogItem>`, eliminating adapter wrappers:
```dart
class Catalog extends core.Catalog<CatalogItem> {
  final List<String> systemPromptFragments;

  const Catalog(
    List<CatalogItem> components, {
    List<ClientFunction> functions = const [],
    required String super.id,
    this.systemPromptFragments = const [],
  }) : super(components: components, functions: functions);
}
```

---

## 4. Key Simplifications & Code Deletions

### 1. Simplify `BasicFunctions`
We can convert all basic functions to pure, synchronous logic returning static values. For example, `AndFunction`:
```dart
class AndFunction extends ClientFunction {
  @override
  Object? execute(Map<String, dynamic> args, core.DataContext context, [...]) {
    final values = args['values'];
    if (values is! List) return false;
    return values.every((v) => _isTruthy(v));
  }
}
```

### 2. Delete redundant parser
`genui/lib/src/functions/format_string.dart` and `ExpressionParser` can be completely removed. We can reuse the `formatString` function already present in `a2ui_core/lib/src/processing/basic_functions.dart` or keep a simplified version returning `ReadonlySignal`.

### 3. Streamlined Widget Data Binding
Widget-binding helpers (e.g., `BoundString`, `BoundBool`) can directly use the signal-backed values from the context since `core.DataContext` provides signal-based resolution (`resolveListenable`).

---

## 5. Migration Strategy

1. **Phase 1: Update Core Interfaces**
   Change `CatalogItem` and `ClientFunction` to implement `core.ComponentApi` and `core.FunctionImplementation`.

2. **Phase 2: Migrate Call Sites & Standard Functions**
   Rewrite the 14 standard functions in `BasicFunctions` to return static values or `ReadonlySignal`s.

3. **Phase 3: Clean up rendering adapters**
   Update `SurfaceController` to pass `Catalog` directly to `core.MessageProcessor` and remove `coreCatalogFor`.

4. **Phase 4: Delete stream parser**
   Delete `ExpressionParser` and the `stream_transform` dependency. Update `BoundValue` widgets to listen to the signal notifier natively provided by the core layer.

jacobsimionato (commented):
Overall, I like this light-touch approach of migrating models to start with and avoiding breaking things.

My main concern is not breaking functionality with each step. I tried the PR out locally, and it did seem like there were some issues with conversation flow etc which I couldn't replicate at head, so wondering if you can look into them before merging?

Do you mind trying out my party planning prompt and seeing if you see the same issues? It could be that my experience is just LLM noise and it's intermittently broken at head already.

Basically, if we can be confident this isn't breaking anything in an obvious way, then I'm keen to merge.

## With PR

### Issue with data model references shown to user

Image

### Issue with conversation flow

The agent is not seeing the item that was chosen in the radio buttons for some reason. It didn't know that it's a graduation party.

Image

## At head

I didn't see these issues, not sure if it's a fluke. The agent did know it's a graduation party though.

Image

jacobsimionato:
> > My main question is whether you have any ideas about avoiding developers needing to add the `core.` import prefixes to resolve symbol conflicts temporarily. Probably there's nothing we can do, but it'd be amazing if we could avoid it.
> 
> I didn't consider this exhaustively. I pretty much always default to the aliasing `import/as` syntax whenever name collisions are involved`–it's more robust when working on frequently changing code and reduces ambiguity when reading code when the reader knows there are many name collisions in the current context[1](#user-content-fn-1-338ec79a7498e1b2a6adb5ea56418fc6).
> 
> > Could we have a separate top-level import file for the conflicting symbols perhaps?
> 
> Unless I'm missing something, isn't this equivalent to consumers switching to the selective `show` import syntax? Regardless, if you have a preference for `show` over `as`, I'm happy to reconsider more thoughtfully.
> 
> ## Footnotes
> 1. Aside : a classic example from my personal experience is from game programming where libraries for physics, 3d rendering, computer vision, spatial audio, etc. all export primitive math types/functions like `Vector3` etc. [↩](#user-content-fnref-1-338ec79a7498e1b2a6adb5ea56418fc6)

Hey no worries, let's leave this, no need to think hard about it. Thanks so much for giving me more context.

andrewkolos:
Thanks for all the additional feedback! I'll be going through each individually, but I can start with the quality regression quality concern since that's also the thing I want to get right here. I actually wasn't able to repro your report here, I experienced failure on both main and here on the 5+ attempts I tried (e.g. I would select "birthday" and the generated text did not acknowledge this in any way and the rest of the generation remained very generic, or I would get into an infinite loop where I kept getting asked for the party type).

At first glance, it doesn't even look like user input gets routed to the LLM at all. The submit button seems to only send up an empty context `{"action":{"name":"submitPartyDetails","context":{}}}`. The client data model isn't being included in the payload.

However, what I am completely confounded by is that when you ran the app on main, the text produced by the agent acknowledges that the party is for a graduation, it sees the date, it sees the number of guests. I don't see how. Regardless, I can fix this in this PR.

As an aside, I notice that earlier text in the conversation seems to be getting retroactively updated. Said another way, the agent is not acting with an append-only approach (or perhaps payloads are malformed). I think simple_chat has been buggy for quite a while, but I digress.

[^1]: https://github.com/flutter/genui/blob/65fab556637c6a1763568f7cb531a55a08eae591/examples/simple_chat/lib/chat_session.dart#L67

jacobsimionato:
Hey thanks for bearing with me with investigating the regressions. As discussed over VC today, it seems like both of the issues I mentioned are actually not regressions and we were able to reproduce them at head. So let's merge this and move on! I can file bugs for the issues.

jacobsimionato (approved):

packages/genui/lib/src/widgets/surface.dart
andrewkolos: I like the thinking, `ActionDelegate.handleEvent`[^1] already exposes `SurfaceDefinition`, so we'd still have to reconstruct that every time to avoid breaking component authoring API. I think per our discussion today, it would be more economical to get through https://github.com/a2ui-project/a2ui/issues/1282 first.

[^1]: https://github.com/flutter/genui/blob/bb50a983c49216348a0de91751585f60e6cbcd6b/packages/genui/lib/src/widgets/surface.dart#L183

packages/genui/lib/src/engine/surface_controller.dart
andrewkolos: Agreed. I Added a TODO to increase our chances of remembering to do this later.

PR #976, opened 2026-06-17
feat(catalog_gallery): bundle samples as assets so the gallery runs on web

polina-c:
contributes to https://github.com/flutter/genui/issues/907

## What

Makes `dev_tools/catalog_gallery` run on the **web** while keeping desktop/mobile working, by bundling the `*.sample` files as Flutter assets and loading them through the asset bundle instead of `dart:io`.

Previously the Samples tab read files via `dart:io` (`Directory.current`, `Directory.list`, `File.readAsString`), which crashes on web at runtime with `Unsupported operation: _Namespace`. The gallery was therefore desktop/mobile only.

## Changes

- **pubspec.yaml** — declare `samples/` under `flutter: assets:` so the sample files ship with the app on every platform.
- **lib/sample_source.dart** (new) — a platform-agnostic `SampleRef` (display name + `Future<String>` loader) and a `SampleSource` abstraction with two implementations:
  - `AssetSampleSource` — enumerates bundled samples via `AssetManifest.loadFromAssetBundle(rootBundle).listAssets()` (keys under `samples/` ending in `.sample`) and reads each with `rootBundle.loadString`. Works on all platforms, including web.
  - `DirectorySampleSource` — desktop/mobile-only filesystem source (`assert(!kIsWeb)`), used for the `--samples` override and tests.
- **lib/samples_view.dart** — depends on `SampleSource`/`SampleRef` instead of `dart:io` `File`/`Directory`; parses content with the existing `SampleParser.parseString`.
- **lib/main.dart** — defaults to `AssetSampleSource` on all platforms. The optional `--samples <dir>` override is guarded behind `!kIsWeb`, so no `dart:io` filesystem call ever runs on web. The old `<cwd>/samples` auto-detection was removed. The Catalog tab is unchanged.
- **README.md** — documents that the gallery runs on web and desktop, shows `flutter run` / `-d chrome` / `-d macos` / `flutter devices`, notes that samples are bundled, and documents `--samples <dir>` as a desktop-only override. Removed the "web is not supported" caveat.
- **tests** — `test/widget_test.dart` and `integration_test/app_test.dart` construct the app with a `DirectorySampleSource`.

## Verification

- `flutter analyze` → No issues found.
- `flutter test` → all passing, including a test that loads samples via `DirectorySampleSource`.
- **Web** (`flutter run -d chrome`) → starts cleanly with no `_Namespace` / `Unsupported operation` / uncaught exception during the automatic startup sample-load; the built web bundle's `AssetManifest` contains all 46 sample keys under `samples/`.
- **Desktop** (`flutter build macos` + `flutter run -d macos`) → builds and launches cleanly; the sample list loads.

https://github.com/user-attachments/assets/156145f9-d516-4041-be41-343fee799e5c


dev_tools/catalog_gallery/lib/main.dart
```diff
@@ -3,47 +3,46 @@
...
+    if (results.wasParsed('samples')) {
+      const FileSystem fs = LocalFileSystem();
+      sampleSource = DirectorySampleSource(
+        fs.directory(results['samples'] as String),
+      );
     }
   }
 
-  runApp(CatalogGalleryApp(samplesDir: samplesDir, fs: fs));
+  runApp(CatalogGalleryApp(sampleSource: sampleSource));
```
ditman, line 35: This is a very good change, thanks!

dev_tools/catalog_gallery/test/widget_test.dart
```diff
@@ -15,15 +16,14 @@ import 'src/test_http_client.dart';
 
 void main() {
   testWidgets('Smoke test', (WidgetTester tester) async {
-    final fs = MemoryFileSystem();
     // Build the app and trigger a frame.
     await tester.pumpWidget(
-      CatalogGalleryApp(fs: fs, splashFactory: NoSplash.splashFactory),
+      const CatalogGalleryApp(splashFactory: NoSplash.splashFactory),
```
ditman, line 21: I wonder if we want a NoopSampleSource or similar to keep this test hermetic (not touching the fs or assets)

dev_tools/catalog_gallery/README.md
```diff
@@ -9,10 +9,28 @@ A developer tool for visualizing and testing the basic widget catalog. Displays
...
+`flutter run -d chrome` (web) or `flutter run -d macos` (desktop). Run
+`flutter devices` to list what is available.
+
+### Loading samples from a custom directory (desktop only)
+
+On desktop/mobile you can override the bundled samples with a directory on
+disk:
+
+```bash
+flutter run -d macos --samples /path/to/samples
```
ditman, line 33: On mobile, would this work on iOS? I'm not sure the iOS app would have access to paths outside its bundle?

dev_tools/catalog_gallery/lib/sample_source.dart
```diff
@@ -0,0 +1,99 @@
...
+
+  /// The display name (the file name without its `.sample` extension).
+  final String name;
+
+  /// Loads the raw contents of the sample.
+  final Future<String> Function() load;
+}
+
+/// A source of [SampleRef]s.
+abstract class SampleSource {
```
ditman, line 27: Should this be an `abstract interface class SampleSource` (instead of an `abstract class`)? at least the way it's being used later makes it sound like an [interface](https://dart.dev/language/class-modifiers#abstract-interface).

ditman (approved):
Small comments, non blocking!

polina-c:
Thank you. You comments makes sense, but I do not want to spend more time for this example now. I want to limit my effort to critical things.

PR #1011, opened 2026-07-23
Node layer for a2ui_core + Flutter

andrewkolos:
Not ready for review. Submitting this early as a draft so it's easier to reference by implementers of other renderers/adapters. The same prototype exists for web_core + React as a2ui-project/a2ui#2077.

(ai-generated; rewrite coming later)

## What it is

The node layer compiles a surface's flat component map plus data model into a live tree of `ComponentNode`s whose resolved properties update reactively. `NodeSurface` ships beside the existing `Surface`, with no changes to it.

## The approach

Nodes are long-lived, each with its own `props` signal. `NodeSurface` wraps each node's subtree in a builder subscribed to that node's `props`, so a data write rebuilds only the node whose resolved properties changed, never the widget tree. The existing `Surface` rebuilds the whole widget tree from an immutable snapshot on every update; `NodeSurface` is the granular sibling, so per-node updates are achievable in a declarative UI framework without whole-surface rebuilds.

## What's here

- **`a2ui_core`**: `ComponentNode` + `NodeResolver`, with the conformance suite ported from web_core (`node_resolver_test.dart`) covering identity, notification granularity, placeholders, templates, and late action resolution.
- **`genui`**: `NodeSurface`, reusing the existing catalog views unchanged, each node subscribed to its own props signal (`node_surface_test.dart`).
- **`examples/simple_chat`**: opts into `NodeSurface` behind `--dart-define=nodes=true` for manual end-to-end runs.


andrewkolos:
Closing since it stalled and I'll be re-implementing the Flutter adapter in the a2ui repo anyway

PR #1017, opened 2026-08-10
Fix chunk newline.

### LINKED ISSUE #1891: Simple chat: error concatenating response.
_↴ Ported from [flutter/genui#906](https://github.com/flutter/genui/issues/906) — originally opened by [polina-c](https://github.com/polina-c) on 2026-05-10._  
_Original labels: front-line-handled_

---

Image

Not sure it is possible to solve, because it may be it is not possible to detect if a space should be added:

Example of inconsistency in chunks:

- "plen"
- "ty of friendly"
- "spots"

polina-c:
Fixes https://github.com/a2ui-project/a2ui/issues/1891 by removing `.trim()` in transport layer.


examples/simple_chat/lib/chat_session.dart
```diff
@@ -121,7 +121,7 @@ sealed class ChatSession extends ChangeNotifier {
       _currentAiMessage = Message(isUser: false, text: '');
       _messages.add(_currentAiMessage!);
     }
-    _currentAiMessage!.text = (_currentAiMessage!.text ?? '') + chunk;
+    _currentAiMessage!.text = '${_currentAiMessage!.text ?? ''}\n$chunk';
```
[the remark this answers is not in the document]
polina-c: by design

jacobsimionato: Are you sure this works as expected?

My understanding is that the LLM emits a sequence of chunks structured however it wants, and there are no guarantees that the chunk boundary coincides with a newline, so therefore we should not inject additional newlines automatically, because they might be in the wrong place.

I don't fully understand the problem you're fixing, but if it's a lack of whitespace in the layout, perhaps this could be accomplished by post-processing the messages at render time to non-destructively add a newline at the end, rather than destructively adding it here in the chunk processing logic.

polina-c: Check screensots and comments on the issue to better understand both issue and fix: https://github.com/a2ui-project/a2ui/issues/1891

polina-c: new line in most cases is noop for markdown, that makes me think the fix is right

jacobsimionato: Hey Polina! Sorry I should have looked at the linked issue first to get more context - I'll do that next time.

I do think that it could cause problems to add a newline here because it's not always a no-op in markdown and could accidentally break words.

## Analysis from Gemini

It turns out the issue isn't that the LLM is omitting spaces between chunks, but rather that [`A2uiTransportAdapter.incomingText`](https://github.com/flutter/genui/blob/main/packages/genui/lib/src/transport/a2ui_transport_adapter.dart#L65-L69) is explicitly stripping them. 

Specifically, in [`packages/genui/lib/src/transport/a2ui_transport_adapter.dart#L68`](https://github.com/flutter/genui/blob/main/packages/genui/lib/src/transport/a2ui_transport_adapter.dart#L68):
```dart
  @override
  Stream<String> get incomingText => _pipeline
      .where((e) => e is TextEvent)
      .cast<TextEvent>()
      .map((e) => e.text.trim())  // <-- Strips leading/trailing whitespace from every chunk!
      .where((text) => text.isNotEmpty);
```

Because .trim() is called on every TextEvent chunk before it reaches _updateAiMessage, any trailing or leading whitespace in streamed chunks (e.g. "ty of friendly ") gets stripped. When _updateAiMessage concatenates these trimmed chunks, words smash together across boundaries (e.g. "friendlyspots").Adding \n before every chunk in chat_session.dart forces literal line breaks inside sentences when chunks split mid-sentence or mid-word, leading to broken streaming rendering and unexpected formatting in Markdown (such as inside code blocks, tables, or lists).

#### Suggested Fix
Instead of appending in chat_session.dart, we should remove `.trim()` from `A2uiTransportAdapter.incomingText` so that original chunk whitespace is preserved as emitted.

polina-c: that's cool! I am impressed.

which model did you use?

applied the fix

jacobsimionato: I think that was just flash actually! I'll send you a screenshot of my workflow :-D

jacobsimionato (approved):

PR #1018, opened 2026-08-11
Reserve name in pub.dev for agent sdk.

polina-c:
https://pub.dev/packages/a2ui_agent

### DIFF
--- a/packages/a2ui_agent/CHANGELOG.md
+++ b/packages/a2ui_agent/CHANGELOG.md
@@ -1,0 +1,3 @@
+## 0.0.1
+
+- Initial version.
--- a/packages/a2ui_agent/CHANGELOG.md
+++ b/packages/a2ui_agent/CHANGELOG.md
@@ -1,3 +1,3 @@
-## 0.0.1
+## 0.0.1-wip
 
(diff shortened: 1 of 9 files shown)


jacobsimionato (approved):
SGTM, though when we actually build this, let's do it in the A2UI repo.

Using "agent" instead of "inference" is what you and @nan-yu agreed on right? I keep forgetting which term we agreed on, but I'm fine with either inference or agent as long as there is consensus.

polina-c:
> SGTM, though when we actually build this, let's do it in the A2UI repo.
> 
> Using "agent" instead of "inference" is what you and @nan-yu agreed on right? I keep forgetting which term we agreed on, but I'm fine with either inference or agent as long as there is consensus.

"you and @nan-yu agreed on" - not yet: before suggesting I want to make sure the name is available. But, while it is '-wip', we can change it.

"when we actually build this, let's do it in the A2UI repo" - I do not think it is good idea - it will be harder, because I need to reference the package from genui. And, it will be tech debt, that, if stay long, will bring even more mess. So, let's coordinate when we actually plan to merge repos.

PR #1020, opened 2026-08-11
Create initial implementation for a2ui_agent.

polina-c:
Contributes to: https://github.com/a2ui-project/a2ui/issues/2356

Prompt:

Use this blueprint to implement the library a2ui_agent: https://github.com/a2ui-project/a2ui/blob/main/blueprints/modules/a2ui_agent.blueprint.md

Next steps:
- test coverage
- use in genui and create migration plan
- port to g3
- migrate users


packages/a2ui_agent/lib/src/inference_formats/direct_json/streaming.dart
```diff
@@ -0,0 +1,401 @@
...
+      final Map<String, dynamic> component = entry.cast<String, dynamic>();
+      final Object? id = component['id'];
+      final Object? type = component['component'];
+      // A component without an id or a type is still arriving; it is not
+      // renderable and must not be emitted as though it were.
+      if (id is! String || id.isEmpty || type is! String || type.isEmpty) {
+        continue;
+      }
+      final String encoded = jsonEncode(component);
+      final key = '$index $id';
```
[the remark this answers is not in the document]
polina-c: obsolete

packages/a2ui_agent/lib/src/utils/catalog_document.dart
```diff
@@ -0,0 +1,313 @@
...
+    final Map<String, dynamic> themeSchema = catalog.themeSchema!.toJsonMap();
+    expandSchemaRefs(themeSchema);
+    theme = themeSchema['properties'] as Map<String, dynamic>?;
+  }
+
+  return {
+    'catalogId': catalog.id,
+    'components': components,
+    if (functions.isNotEmpty) 'functions': functions,
+    'theme': ?theme,
```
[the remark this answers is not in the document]
polina-c: it is valid

see https://dart.dev/language/collections#null-aware-elements

polina-c:
If looks ok, I will resolve circlular dependencies.

packages/a2ui_agent/lib/src/parser/incremental_processor.dart
```diff
@@ -0,0 +1,110 @@
...
+import 'response_part.dart';
+import 'sentinel_tokenizer.dart';
+
+/// Shared streaming plumbing for format parsers.
+///
+/// The processor owns the sentinel tokenizer and the raw buffer for the block
+/// currently being streamed. Formats only implement [emitDelta], which is
+/// handed the whole raw block accumulated so far and returns the messages that
+/// have not been emitted yet.
+abstract class IncrementalStreamProcessor {
```
jacobsimionato, line 14: This seems like a useful utility - I wonder if it should be in the shared design!

packages/a2ui_agent/lib/src/utils/catalog_document.dart
```diff
@@ -0,0 +1,317 @@
...
+
+  JsonComponentApi({required this.name, required this.schema});
+}
+
+/// A catalog function known only by its declaration.
+///
+/// Client functions execute on the renderer, so the agent side carries the
+/// signature but cannot run it. [execute] throws to make an accidental
+/// server-side invocation loud instead of silently wrong.
+class DeclaredFunction extends FunctionImplementation {
```
jacobsimionato, line 30: I think we should instead handle FunctionApi and FunctionImplementation similar to how we handle ComponentApi and ComponentImplementation in the core library.

E.g. `FunctionApi` does not declare `execute` at all, and `FunctionImplementation` extends `FunctionApi`.

Then we have `Catalog<C extends ComponentApi, F extends FunctionApi>` perhaps.

packages/a2ui_agent/lib/src/inference_formats/express/ast.dart
jacobsimionato, line 1: @gspencergoog used ANTLR to generate the core parts of this for Python and Kotlin, based on some language definition file. Can we do that for Dart too? It seems supported - https://github.com/antlr/antlr4

See https://github.com/a2ui-project/a2ui/tree/main/agent_sdks/python/a2ui_agent/src/a2ui/inference_formats/experimental/express/generated

jacobsimionato: Also, feel free to omit the express stuff from the initial PR if you like - it's easy to add as a followup because it's behind a well-defined narrow interface (InferenceFormatFactory, InferenceFormat).

packages/a2ui_agent/lib/src/utils/catalog_document.dart
```diff
@@ -0,0 +1,317 @@
...
+    'theme': ?theme,
+  };
+}
+
+/// Builds a [Catalog] from an A2UI catalog [document].
+///
+/// [protocolVersion] and [catalogId], when given, must agree with what the
+/// document declares; a conflict throws [A2uiValidationError] rather than
+/// silently loading a catalog the renderer did not ask for.
+Catalog<ComponentApi> catalogFromDocument(
```
jacobsimionato, line 122: I think the protocol version and catalog Id should be declared in the document anyway, right?

I think this should be implemented in the core library, possibly on the Catalog and Component classes / interfaces themselves.

packages/a2ui_agent/lib/src/prompt/generator.dart
```diff
@@ -0,0 +1,33 @@
...
+
+import 'package:a2ui_core/a2ui_core.dart';
+
+import '../parser/response_part.dart';
+
+/// A named set of example messages used as a few-shot prompt turn.
+///
+/// The key of the surrounding map describes the example turn; the value is the
+/// A2UI payload the model is expected to produce for it.
+typedef PromptExamples = Map<String, List<AgentToRendererMessage>>;
```
jacobsimionato, line 13: I like it!

packages/a2ui_agent/lib/src/prompt/generator.dart
```diff
@@ -0,0 +1,33 @@
...
+/// active catalogs. Callers prepend their own role and workflow preambles and
+/// append their own suffixes.
+abstract class PromptGenerator {
+  /// The active catalogs to describe in the system instructions.
+  final List<Catalog<ComponentApi>> catalogs;
+
+  /// Optional few-shot example turns, keyed by a description of the turn.
+  final PromptExamples? examples;
+
+  const PromptGenerator(this.catalogs, {this.examples});
```
jacobsimionato, line 28: I can't remember what is in the standard design, but I'm wondering if we maybe don't need to expose the catalogs or examples or define a constructor here. Instead, we just expose `generate()` and each inference format can implement this however it wants.

packages/a2ui_agent/lib/src/inference_formats/direct_json/prompt_generator.dart
jacobsimionato, line 1: @nan-yu if we don't have them already, let's add conformance tests requiring that each prompt generator implementation returns *exactly* the same prompt for a given catalog. Then we know that the eval work we do will translate across library implementations.

jacobsimionato (commented):
Hey this is looking great - nice work Polina and Gemini! It is a lot of code!

I'm actually finding it okay to review, because it's  mostly about the interfaces. But I do wonder about trying to split it a little... e.g. you could probably omit the express format completely in the first PR for example.

I like Nan's idea of trying to do conformance tests up front too. Though they will only really test the implementation, and the more interesting part to review as a human is the APIs and how you've translated them to Dart. So maybe we can human-review the APIs here and not worry too much about the implementation, then add conformance tests and have an agent fix the implementations where necessary.

I dunno, curious about @nan-yu 's thoughts on the best process also. I'd like to make sure that both of us get a proper chance to review the APIs here, because they are important and difficult to change later.

jacobsimionato:
Also, I think we reached consensus to do this in the A2uI repo, right? I know that was after you created this PR though!

polina-c:
> Also, I think we reached consensus to do this in the A2uI repo, right? I know that was after you created this PR though!

I am confused.

I thought we have agreement to copy it to a2ui repo after a2ui repo is ready, together with a2ui_core, that is in monorepo with this code (see https://github.com/a2ui-project/a2ui/issues/2234). 

Do you suggest to hold development?

packages/a2ui_agent/lib/src/processor/renderer_capabilities.dart
nan-yu, line 1: This should belong to the Dart a2ui_core library, which maps the web_core version, https://github.com/a2ui-project/a2ui/blob/main/renderers/web_core/src/v0_9/schema/client-capabilities.ts#L55.

It actually brings up a broader issue. Should we update the Dart a2ui_core library first, and then build the a2ui_agent on the latest a2ui_core?

packages/a2ui_agent/lib/src/validation/payload_validator.dart
nan-yu, line 1: The validation layer should also reside in the a2ui_core layer.

packages/a2ui_agent/lib/src/inference_formats/direct_json/prompt_generator.dart
nan-yu: We have it already in https://github.com/a2ui-project/a2ui/blob/a70e0ebc7ec37d848f1bd6194d10cc135efb5ab0/agent_sdks/conformance/suites/inference_format.yaml#L191.

Just a heads up, we're planning to restructure the conformance suite to be more modularized, but it will have same coverage.

packages/a2ui_agent/lib/src/processor/catalog_providers.dart
```diff
@@ -0,0 +1,142 @@
...
+  /// The protocol version to load the bundled catalog for.
+  final ProtocolVersion protocolVersion;
+
+  const BundledCatalogProvider({
+    this.protocolVersion = ProtocolVersion.current,
+  });
+
+  @override
+  Catalog<ComponentApi> load() {
+    if (protocolVersion == ProtocolVersion.v09) return MinimalCatalog();
```
nan-yu, line 36: Why MinimalCatalog? Do we have a BasicCatalog in Dart?

packages/a2ui_agent/lib/src/processor/catalog_providers.dart
```diff
@@ -0,0 +1,142 @@
...
+  /// Loads the catalog.
+  Catalog<ComponentApi> load();
+}
+
+/// Loads a catalog bundled with the SDK for a protocol version.
+///
+/// `package:a2ui_core` bundles the minimal catalog for `v0.9`. Other versions
+/// have no bundled catalog in Dart yet, so asking for one is an error rather
+/// than a silent fallback to a catalog the renderer never agreed to.
+class BundledCatalogProvider extends CatalogProvider {
```
nan-yu, line 26: I forgot to update the blueprint when we decided to remove the BundledCatalogProvider class. Nothing needs to be bundled with the SDK. We can use the BasicCatalog directly from a2ui_core. I'll update the blueprint separately.

packages/a2ui_agent/lib/src/processor/processor.dart
```diff
@@ -0,0 +1,105 @@
...
+  String get promptSnippet =>
+      _promptSnippet ??= format.promptGenerator.generate();
+
+  /// Creates a parser for one model turn.
+  ///
+  /// Parsers accumulate streaming state, so a turn must not share one with
+  /// another turn.
+  Parser createParser() => format.createParser();
+
+  /// Parses and validates a complete model response.
```
nan-yu, line 61: The validation part is missing here.

packages/a2ui_agent/lib/src/utils/catalog_document.dart
nan-yu: This should also be moved to the a2ui_core layer.

packages/a2ui_agent/lib/src/utils/catalog_document.dart
nan-yu: Yes, the Catalog class is designed to have  `from_json` and `catalog_schema`, which builds a Catalog class from a JSON schema, and returns the catalog schema in JSON representation.

packages/a2ui_agent/lib/src/validation/payload_validator.dart
```diff
@@ -0,0 +1,414 @@
...
+
+import 'package:a2ui_core/a2ui_core.dart';
+import 'package:json_schema_builder/json_schema_builder.dart';
+
+import '../parser/response_part.dart';
+import '../primitives/protocol_version.dart';
+import '../utils/schema_utils.dart';
+
+/// A single problem found in an A2UI payload.
+class A2uiValidationIssue {
```
nan-yu, line 13: Can we rename to A2uiValidationError? It should also go to the a2ui_core layer.

packages/a2ui_agent/lib/src/validation/payload_validator.dart
```diff
@@ -0,0 +1,414 @@
...
+    return location.isEmpty ? message : '$message (in ${location.join(', ')})';
+  }
+}
+
+/// Validates compiled A2UI payloads against the negotiated catalogs.
+///
+/// This is the agent SDK's validation layer. The A2UI agent specification
+/// delegates it to `A2uiValidator` in the core package;
+/// `package:a2ui_core` does not ship one yet, so the checks live here and
+/// operate on the `v0.9` envelopes that package models.
```
nan-yu, line 40: The validator in a2ui_core should support validation against all protocol versions.

packages/a2ui_agent/test/catalog_document_test.dart
nan-yu, line 1: Most of these unit tests should be moved to conformance spec, so we don't repeat for each implementation.

polina-c:
Thank you for the comments. I plan to address most of them in blueprint and a2ui_core and regenerate from scratch.

PR #1022, opened 2026-08-12
Fix and stabilize TextField.

polina-c:
Contributes to https://github.com/a2ui-project/a2ui/issues/1893

1. Update examples to be more helpful in explaining how to use the component TextField.
2. Fix password example to be indeed obscured.
3. Move string constants to constants to avoid error above in future
4. Fix component implementation by type:
- obscured should be obscured
- longText should support multiple lines
- number should support only numbers
5. Linked the package in CHANGELOGs to simplify maintanance.
6. Implemented validationRegexp.

Screenshot 2026-08-11 at 7 45 14 PM

### DIFF
--- a/packages/genui/lib/src/catalog/basic_catalog_widgets/text_field.dart
+++ b/packages/genui/lib/src/catalog/basic_catalog_widgets/text_field.dart
@@ -191,8 +191,8 @@ final textField = CatalogItem(
         {
           "id": "root",
           "component": "TextField",
-          "value": "Hello World",
-          "label": "Greeting"
+          "label": "Enter your name here:"
+          "textFieldType": "shortText"
         }
       ]
     ''',
@@ -201,8 +201,38 @@ final textField = CatalogItem(
         {
           "id": "root",
           "component": "TextField",
-          "value": "password123",
-          "label": "Password",
+          "label": "Type your story here:",
+          "textFieldType": "longText"
+        }
+      ]
+    ''',
+    () => '''
+      [
+        {
+          "id": "root",
+          "component": "TextField",
+          "label": "Type your story here:",
+          "textFieldType": "longText"
+          "value": "Once upon a time..."
+        }
+      ]
+    ''',
+    () => '''
+      [
+        {
+          "id": "root",
+          "component": "TextField",
+          "label": "What is your initial price?",
+          "textFieldType": "number",
+        }
+      ]
+    ''',
+    () => '''
+      [
+        {
+          "id": "root",
+          "component": "TextField",
+          "label": "Enter your password here",
           "textFieldType": "obscured"
         }
       ]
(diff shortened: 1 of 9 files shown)


packages/genui/lib/src/catalog/basic_catalog_widgets/text_field.dart
```diff
@@ -5,29 +5,61 @@
...
+    _Json.value: A2uiSchemas.stringReference(
       description: 'The value of the text field.',
     ),
-    'label': A2uiSchemas.stringReference(),
-    'variant': S.string(
-      enumValues: ['shortText', 'longText', 'number', 'obscured'],
+    _Json.label: A2uiSchemas.stringReference(),
+    _Json.variant: S.string(
+      description:
+          '''The kind of input the field accepts. ${_Variant.shortText} (the default) is a single line of text, ${_Variant.longText} is multi-line text, ${_Variant.number} only accepts numeric input, and ${_Variant.obscured} hides the typed characters, e.g. for passwords.''',
```
jacobsimionato, line 47: I think we should standardize these descriptions to exactly match central catalog.json document e.g. https://github.com/a2ui-project/a2ui/blob/main/specification/v0_9/catalogs/basic/catalog.json

That way, rather than iterating on the descriptions separately in every codebase, we can iterate on them in one central place, have central evals that show that they work well, and then just copy them everywhere.

So would it work to just copy the description from the central catalog doc e.g. "The type of input field to display."? And if that causes inference performance issues, then send a PR to the JSON file for the basic catalog for 1.0 to improve the description.

packages/genui/lib/src/catalog/basic_catalog_widgets/text_field.dart
```diff
@@ -5,29 +5,61 @@
...
 import '../../model/a2ui_schemas.dart';
 import '../../model/catalog_item.dart';
 import '../../model/data_model.dart';
 import '../../model/ui_models.dart';
 import '../../model/validation_helper.dart';
+import '../../primitives/logging.dart';
 import '../../primitives/simple_items.dart';
 import '../../widgets/widget_utilities.dart';
 
+class _Json {
```
jacobsimionato: Optional bikeshed: How about `_Fields` e.g. `_Fields.value`? The name _JSON to me suggests we're calling a JSON parsing library of some kind.

jacobsimionato (approved):

packages/genui/lib/src/catalog/basic_catalog_widgets/text_field.dart
polina-c: I like this.
But I suggest it to be separate effort. 
I do not want to spend more time here now.
I just wanted to try to fix https://github.com/a2ui-project/a2ui/issues/1893, plus fixed what I saw was not working.

Would you mind to file separate issue for description standartization?

gspencergoog: Really, we should be using some tooling to build these schema from the A2UI schema files themselves.  They should be *identical*, not just copied over: copied ones will always be stale.

PR #1026, opened 2026-08-20
Fix/email validation regex

cubegao:
Fix the email validation function by correcting the regex escape pattern.

The previous implementation used an escaped `$` in a raw string, which caused the regex to match a literal `$` character instead of the end-of-string anchor.

Update the regex pattern to properly validate email formats.

# Description

This PR fixes an issue in the email validation function where the regular expression pattern did not correctly handle the end-of-string anchor.

The previous implementation incorrectly escaped the `$` character inside a Dart raw string. As a result, the regex engine treated it as a literal character instead of the end-of-string matcher, causing valid email addresses to fail validation.

This change updates the regex pattern to use the correct `$` anchor and ensures email validation behaves as expected.

## Related Issues

N/A

## Testing

- Verified that valid email addresses are correctly accepted.
- Verified that invalid email formats are rejected.

## Pre-launch Checklist

- [x] The code changes have been tested.
- [x] The updated validation logic works as expected.
- [ ] Documentation updates are not required.

### DIFF
--- a/packages/genui/lib/src/catalog/basic_functions.dart
+++ b/packages/genui/lib/src/catalog/basic_functions.dart
@@ -309,7 +309,7 @@ class EmailFunction extends SynchronousClientFunction {
   Object? executeSync(JsonMap args, ExecutionContext _) {
     final Object? value = args['value'];
     if (value is! String) return false;
-    final emailRegex = RegExp(r'^[^@]+@[^@]+\.[^@]+\$');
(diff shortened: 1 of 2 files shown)


josemontespg:
@gspencergoog could you review this one? It is a one-character fix: in the raw string the "\$" matched a literal dollar sign, so the email function rejected every real address; the PR adds tests. The Flutter GenUI CI run for this PR never started (0 jobs) and needs a re-run before merging.

